PR #19, opened 2016-03-19
add parser for canteen belonging to erlangen nuernber

theScrabi:
Hello,

I've created a parser so canteens belonging to "Studentenwerk Erlangen Nürnberg" get available via openmensa.org.

rgds Schabi

### DIFF
--- a/erlangen_nuernberg.py
+++ b/erlangen_nuernberg.py
@@ -38,8 +38,11 @@ def parse_url(url, today=False):
                 return False
         return True
 
+    def is_action_entry(td):
+        return td.text == 'Aktion'
+
     def is_closed(tds):
-        return is_new_entry(tds) and get_pricing(tds) is None
+        return is_new_entry(tds) and get_pricing(tds, 4, 7) is None
 
     def refactor_date(raw_date):
         now = datetime.datetime.now()
@@ -60,7 +63,7 @@ def parse_url(url, today=False):
             img = td.find_all('img')[0]
             src = img.get('src')
             if('msc' in src):
-                type += 'MSC Fish '
+                type += 'Fish MSC '
             elif('vegan' in src):
                 type += 'Vegan '
         else:
@@ -83,7 +86,6 @@ def parse_url(url, today=False):
     def get_refs(td):
         return td.find_all('sup')
 
-
     def get_foot_description(td):
         refl = get_refs(td)
         description = td.text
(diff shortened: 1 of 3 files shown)


theScrabi:
Also a small question:
When I add a new source/canteen on the website.
What should I enter in the "Name" and "Meta-URL" field?

erlangen_nuernberg.py
```diff
@@ -0,0 +1,234 @@
...
+            elif r == 'a11':
+                food_contains += 'Sesam, '
+            elif r == 'a12':
+                food_contains += 'Schwefeldioxid/Sulfite, '
+            elif r == 'a13':
+                food_contains += 'Lupinen, '
+            elif r == 'a14':
+                food_contains += 'Weichtiere, '
+            else:
+                food_contains += 'undefinierte Chemikalien:'+r+', '
```
mswart: Could you extract this data from the website - so updates would be found automatically?

erlangen_nuernberg.py
```diff
@@ -0,0 +1,234 @@
...
+                type += 'Geflügel '
+            if('V' in td.string):
+                type += 'Vegetarisch '
+            if('F' in td.string):
+                type += 'Fisch '
+            if('L' in td.string):
+                type += 'Lamm '
+            if('W' in td.string):
+                type += 'Wild '
+        return type
```
mswart: Why a space at the end?

mswart:
The name should be a short description of the canteen, like "Mensa X" - sometimes with a city name.

The meta URL is a way to provide meta-information (like address and the feeds) about a canteen directly from the parser - with the option to parse the information from the upstream website. It is currently not really used (you can take a look at the parser for `halle`). And documentation is currenlty minimalist (beside the feed v2 description).

erlangen_nuernberg.py
theScrabi: Because you can have combinations off food. So There can bee a "Beef Pork" topic. I know "Beef Prok " is not that nice. If it causes trouble I'll remove that space at the end.

erlangen_nuernberg.py
theScrabi: That would be tough work for a small benefit. I had a conversation with the people providing the website. According to them these descriptions almost never change.

theScrabi:
So if I do get it right, Meta URL should be an URL similar to the [shared_prefix](https://github.com/mswart/openmensa-parsers/pull/19/files#diff-37cd9ad3426e6988ea7a0cf4bb4beae0R222) URL combined with one of its suffix?

mswart:
`shared_prefix` and the Meta URL are two very different things. The `shared_prefix` is a parser internal option/shortcut to simplify generating the upstream URL that should be parsed.

The Meta URL is a optional setting for OpenMensa canteen and defines a URL to should be fetch to get meta data about the canteen.

Lets take the `Heidenmensa` in Halle as example (https://openmensa.org/c/238):

* **http://omfeeds.devtation.de/halle/heidemensa/metadata.xml**: this is the meta-URL for the canteen 238. The URL is fetched once a day (currently ~4 am). Although it returns meal information these data are ignored. The provided canteen meta data like address and opening times are used to inform the author about changes (they are currently not applied automatically). In addition multiple feeds are defined. The feed is a URL returning meal-information and a schedule info.
* **http://omfeeds.devtation.de/halle/heidemensa/today.xml**: This is the smallest feed the only parses the current date and it is request hourly during the day so that changes are detected fast by OpenMensa.
* **http://omfeeds.devtation.de/halle/heidemensa/thisWeek.xml** and **http://omfeeds.devtation.de/halle/heidemensa/nextWeek.xml**: are the base feeds that are used to fetch data about the future. They are only fetched once a day, as they are not that important and changes are less likely.

The metadata URL is optional, you can submit the needed information also manual via the OpenMensa website (if you find an issue please report it).

In addition a Canteen must have at least one feed providing the meal information. It is recommended to split the data into multiple feeds but not required. I strongly advice to define a today feed that is fetched frequently to be up-date without querying to much. The feed can include also meal information for other dates that today if they can only be fetched together. To be more resistant against errors I added to separated feeds per week for halle (if you week fails due some day-specific data issue, the other week can still be fetched). But this is a advanced optimization.

If you do not have implemented extracting metadata - I should leave the Meta-URL empty, afterwards you should be able to provided the needed information manual.

If you want to implement extracting  metadata, you should take a look the the utils module with the EasySource as used by the parser of halle.

mswart:
The parser is deployed (e.g. http://omfeeds.devtation.de/erlangen_nuernberg/er-langemarck/full.xml). If you do adjustments/changes please open a new PR.

But as I am on vacation the next week I am not sure how quickly I can review/merge them.

theScrabi:
Ok so I still didn't figure out how to add a new feed to a certain canteen. What format should "Wiederholungsinterval(e)" have, and what should I enter for "Abrufplan"?

theScrabi:
What should I do to make my canteen visible at the openmensa website?

mswart:
Sorry, the documentation is very short on this points.
Please read / take a look at existing on this [part of the feed document](http://doc.openmensa.org/feed/v2/#optional-schedule-element).

"Abrufplan" is the space concatenated  of `minute`, `hour`, `month`, `dayOfWeek`, `dayOfMonth`. In most cases something like `0 8 * * *` (fetch every day at 8 o'clock). If you want to find further infos you can search for cron format.

`Wiederholungsinterval(e)` are space concatenated numbers. Empty means no retry, other examples are given in the link.

The parser/canteen should be visible if the was once successfully fetch (working parser with feed and some data). Until then the canteen/feed is only directly accessible. If this does not work for you, please ping me again so that I can have a look.

PR #22, opened 2016-04-13
Fixed serveral issues

theScrabi:
Yet another pull request had to be done. Sorry for annoying you again. How can I test if openmensa.org will fail or not?

The following changes on erlangen_nuernberg where made:
* add support for "Tipp des Tages"
* made parser support dates in the formate "dd.mm"

### DIFF
--- a/erlangen_nuernberg.py
+++ b/erlangen_nuernberg.py
@@ -8,14 +8,14 @@ from pyopenmensa.feed import LazyBuilder
 
 from utils import Parser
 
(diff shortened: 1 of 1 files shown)


mswart:
You can execute the feed locally by calling something like:
 `python3 parse.py erlangen_nuernberg er-langemarck full.xml`.
It will print the returing xml or raise an exception.

For faster iteration cycles if you add you as collaborator, so you can merge you MR directly. CI + Auto-Deployment normally takes 5-10 minutes to deploy.

theScrabi:
Hei thanks :) Now it works.

PR #28, opened 2016-06-12
chemnitz-zwickau: add parser

a-andre:


theScrabi:
I don't know if I'm permitted to merge requests form other people. ... But I'll do it anyway. If something fails we can still do a rebase.

PR #30, opened 2016-06-13
chemnitz_zwickau: fix price parsing issue

a-andre:
The initial version had some issue with the price. These should be fixed now.


mswart:
Sorry for the late response, had couple of busy weeks.

PR #31, opened 2016-06-27
Fix the leipzig parser

klemens:
The Studentenwerk Leipzig recently launched a new website that broke the old parser, which relied on an ajax api that was used in the old website. The new one offers a dedicated [xml API](https://www.studentenwerk-leipzig.de/XMLInterface/request).

I also fixed a problem of the previous parser which added every component of a meal as a separate meal.

Note that I used BeautifulSoup for parsing the xml, so `lxml` is needed. It seems the debian package already specifies this requirement, but it was missing from the travis config.

### DIFF
--- a/leipzig.py
+++ b/leipzig.py
@@ -1,38 +1,55 @@
 #!python3
 from urllib.request import urlopen
-import json
+from bs4 import BeautifulSoup
 import datetime
 
 from utils import Parser
 
 from pyopenmensa.feed import LazyBuilder
 
+def parse_prices(prices):
+    price_map = {}
+    for price in prices:
+        if '0' == price['consumerID']:
+            price_map['student'] = price.getText()
+        elif '1' == price['consumerID']:
+            price_map['employee'] = price.getText()
+        elif '2' == price['consumerID']:
+            price_map['other'] = price.getText()
+    return price_map
 
-def correct_prices(v):
-    if 'employe' in v:
-        v['employee'] = v.pop('employe')
-    if 'guest' in v:
-        v['other'] = v.pop('guest')
-    return v
-
-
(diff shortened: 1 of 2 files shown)


klemens:
Travis fails due to `http://localhost:9090/darmstadt/stadtmitte/full.xml` returning a `500`, while the new leipzig parser seems to work fine: :wink: 
```text
--2016-06-27 23:07:25--  http://localhost:9090/leipzig/liebigstrasse/full.xml
Connecting to localhost (localhost)|127.0.0.1|:9090... connected.
HTTP request sent, awaiting response... 200 OK
Length: 59400 (58K) [application/xml]
Saving to: ‘/dev/null’

/dev/null           100%[=====================>]  58.01K  --.-KB/s   in 0s     
```

mswart:
Many thx for the fix/rewrite of the parser, I do not have enough time to maintain all parsers closely.

If are interested, you could take over the official maintenance (which mostly means to get emails of the parsers returns unexpected results like 500er). Simple contact me.

mswart:
The parser was used mostly sucessfully - two minor issues:
* peterssteinweg stills throws exceptions
* a couple of feeds (`am-park/full.xml`, `tierklinik/full.xml`, `academica/full.xml`, `am-elsterbecken/full.xml`, `am-elsterbecken/today.xml`) contains empty notes which is not valid coursing openmensa to reject the feed.

Could you take a look at these issues?

klemens:
The issues should be fixed with #32.

 > If are interested, you could take over the official maintenance (which mostly means to get emails of the parsers returns unexpected results like 500er). Simple contact me.

I am interested. :wink: I will contact you next week.

PR #32, opened 2016-06-28
Additional fixes for the leipzig parser

klemens:
Includes a fix for the exception in the `peterssteinweg` feed and avoids generating empty notes.

I have also simplified the fetch logic to always fetch 3 weeks in advance (see corresponding commit).


mswart:
Thx

PR #34, opened 2016-08-29
Update niederbayern-oberpfalz.py

shad0w73:
I fixed the issue "empty note", simplified the code a bit, added comments and updated the usable locations

### DIFF
--- a/niederbayern_oberpfalz.py
+++ b/niederbayern_oberpfalz.py
@@ -53,7 +53,7 @@
 
 import sys
 from csv import reader
-from datetime import date
+from datetime import date, timedelta
 from urllib.request import urlopen
 from urllib.error import HTTPError
 import re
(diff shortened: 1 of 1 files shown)


niederbayern_oberpfalz.py
```diff
@@ -111,88 +110,113 @@ def parse_url(url, today=False):
...
+            if bpos == len(mname)-1:
+                # Remove brackets and break bracket loop
+                mname = mname[:apos]
+                bpos = -1
             else:
-                raise e
-        f = f.read().decode('iso8859-1')
+                # Remove current brackets, improve readability
+                # and find the next brackets
+                mname = mname[:apos] + ' und ' + mname[bpos+1:]
```
klemens: NIT: This generally results in the name containing too many spaces, eg `Jägerschnitzel (3,A,G,I)Bio Schwäbische Spätzle (A,C)` from [UNI-R/37.csv](http://www.stwno.de/infomax/daten-extern/csv/UNI-R/37.csv) results in two spaces between `Jägerschnitzel` and `und Bio Schwäbische Spätzle`.
I would suggest something like `mname[:apos].rstrip() + ' und ' + mname[bpos+1:].lstrip()`.

niederbayern_oberpfalz.py
```diff
@@ -111,88 +110,113 @@ def parse_url(url, today=False):
...
     }
-    #canteen.setLegendData(legend)
 
+    # Create regular expressions for categories
     hg = re.compile("^HG[1-9]$")
     b = re.compile("^B[1-9]$")
     n = re.compile("^N[1-9]$")
 
-    #for w in 0, 1:
-    for w in [0]:
```
klemens: Maybe you could keep the loop and fetch the menu for the current and the next week?

klemens:
Generally this looks good to me (apart from the nits). You should however rebase your branch onto the current master to remove all the merge commits: `git rebase 34f0617`.
I would also recommend including the name of the parser in the commit messages to avoid confusion, something like: `niederbayern: Commented the code`.

niederbayern_oberpfalz.py
shad0w73: I thought about it but then I removed it.
I will add it again because this way it should be better. Especially on monday morning.

niederbayern_oberpfalz.py
shad0w73: Good hint, thank you.

klemens:
Looks good to me now.
I am not sure if the new canteens added in a90afd695379b81d791d7b5277ecf77832f0ed13 are added automatically as sources for https://openmensa.org/ or if they have to be added manually. @mswart?

shad0w73:
For now I will just add WZ Straubing and wait for the new semester to add the remaining locations.

klemens:
Sounds good.
I've added a small additional fix for this parser, which I hadn't thought of before: only fetch the current week when the today feed is requested: 58c330e01cd143c4fd70742185b950e88eae41db

mswart:
Sorry for the late response, I am currently busing moving flats.
@klemens There is a mechanism to fetch list of sources automatically, but it is rarely testing, and not enabled for most canteens. You could try to define the parser's Index-URL (`http://omfeeds.devtation.de/niederbayern_oberpfalz/index.json` in this case) and run "Neue Quelle/Mensa hinzufügen". If you find issues, please report them to the `openmensa/openmensa` project. But it will take same weeks before I can fix them.

Issue #37, opened 2016-10-18
München & Nürnberg do not work anymore


flippinger:
all locations in München/Garching and Nürnberg/Erlangen that I tested do not work anymore.

For example:
- http://openmensa.org/c/256
- http://openmensa.org/c/138
- http://openmensa.org/c/139

API returns error 500.

klemens:
München should work fine now with 748ee85ca465aec889acbcb28678a5fd0a4a6807.

It seems nothing is wrong with Nürnberg except that the menu was last updates at 28.09.2016? I am not sure whats the problem here.. @mswart @theScrabi

mswart:
By df6062c0f209a8bd2436a25a25e957761cde2ae1 and 07346ce769b59a9fe493492733f46ecdfb317bf3 Nürnberg/Erlangen produce again valid XML-Feeds. Update worked fine (new weeks were imported).

flippinger:
Thanks for the quick fix. Nice work!

PR #40, opened 2016-10-19
WIP: Validate documents against the openmensa schema

klemens:
This is useful for development, as you may not notice when the generated xml is not valid (as I did in #37).

The first commit adds validating support to `parse.py`, which makes lxml a mandatory dependency for dev.

I also think it would be useful to add validation to the travis builds, maybe by replacing the wget commands with a python script that validates the returned documents. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ As many meal information are only available online as HTML, [Beautiful Soup 4](h
 2. Install the dependecies:
    * [Python 3](https://www.python.org/)
    * [Beautiful Soup 4](http://www.crummy.com/software/BeautifulSoup/) - needed for most parsers/providers.
-   * [python-lxml](http://lxml.de/) Some parsers using the `lxml` backend of Beautiful Soup, so you might need the Python `lxml` module/extension.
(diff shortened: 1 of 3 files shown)


mswart:
Looks good as fallback layer. I am unsure whether it is good to have the schema copied into this project. Independent from that I have updated `PyOpenMensa` (c3fd2eca4a7569fc825f9bd9ab1eb0d3c6b4576e) to raise `ValueError`s on empty category names and empty notes: mswart/pyopenmensa@05d9fd6c29000dd0246afd2bbf9731caeef3adeb.

For now I would keep wget on travis. My goal would be to adjust `PyOpenMensa` to prevent all invalid XML generation.

klemens:
You are right, adding the schema to the repository isn't so great. I added it, because downloading it on every call to `parse.py` seemed even worse. However making sure `PyOpenMensa` produces valid documents is the far better option.

I think validating on travis would still be useful, especially as you are using it to directly deploy on a successful build. Instead of using a python script for validation we could also just pipe the files through xmllint and download the schema once for every build, so we don't have to add it to the repo.

PR #42, opened 2016-11-04
Aachen: add 'Hauptbeilage' and 'Gemüse/Salat'

manorom:
Currently the 'extras' section of the menu is not parsed. The commit adds code to do that and adds the category 'Hauptbeilage' and 'Gemüse / Salad' to most aachen canteens.

### DIFF
--- a/aachen.py
+++ b/aachen.py
@@ -33,6 +33,25 @@ def parse_day(canteen, day, data):
         # store data
         canteen.addMeal(day, category, name, notes, price)
 
+    # 2. extras:
+    if not data.find(attrs={'class': 'extras'}):
+        return
+    for extra in data.find(attrs={'class': 'extras'}).find_all('tr'):
+        category = extra.find('span', attrs={'class': 'menue-category'}).text.strip()
+        name = ''
+        notes = set()
+        for namePart in extra.find('span', attrs={'class': 'menue-desc'}).children:
+            if type(namePart) is NavigableString:
+                name += namePart.string
+            elif type(namePart) is Tag:
+                if 'or' in namePart.get('class', []):
+                    name += namePart.string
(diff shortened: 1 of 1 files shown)


aachen.py
```diff
@@ -33,6 +33,25 @@ def parse_day(canteen, day, data):
         # store data
         canteen.addMeal(day, category, name, notes, price)
 
+    # 2. extras:
+    if not data.find(attrs={'class': 'extras'}):
+        return
+    for extra in data.find(attrs={'class': 'extras'}).find_all('tr'):
```
klemens: You are searching for the same table twice.

aachen.py
```diff
@@ -33,6 +33,25 @@ def parse_day(canteen, day, data):
...
+        name = ''
+        notes = set()
+        for namePart in extra.find('span', attrs={'class': 'menue-desc'}).children:
+            if type(namePart) is NavigableString:
+                name += namePart.string
+            elif type(namePart) is Tag:
+                if 'or' in namePart.get('class', []):
+                    name += namePart.string
+                else:
+                    notes.update(namePart.text.strip().split(','))
```
klemens: Wouldn't it be better to explicitly check if the tag is a `sup` and append [`.string`](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#string) to the name in all other cases?
This would also work for the normal menus in part 1 and make the code nearly identical, so you could then combine it into one function.

manorom:
Sorry, my first commit was rather quick fix. I implemented your recommendations now.

klemens:
I squashed and merged your commits (bc65b47335d7253187791c672d7782d6280ef73d) and made up a nice branch name. 😉 Thank you!

PR #45, opened 2017-01-04
added parser for "Studentenwerk Düsseldorf"

Athemis:
Parser for canteens operated by "Studentenwerk Düsseldorf". Food additive list is maintained as PDF and difficult to parse, hence a static dictionary compiled from that PDF is included with the parser.

### DIFF
--- a/duesseldorf.py
+++ b/duesseldorf.py
@@ -1,0 +1,163 @@
+import re
+from urllib.request import urlopen
+
+from bs4 import BeautifulSoup as parse
+
+from pyopenmensa.feed import OpenMensaCanteen
+from utils import Parser
+
+# this will match notes, e.g. "[R]" or "(1,10,12)"
+legend_regex = re.compile('[\[(][ARFSVZ\d\,]+[\])]')
+
+# this list was compiled from pdf available at http://www.stw-d.de/wp-content/uploads/2017/01/Zeichenerklaerung.pdf
+legend_dict = {'R': 'mit Rindfleisch',
+               'S': 'mit Schweinefleisch',
+               'A': 'mit Alkohol',
+               'F': 'fleischlos (enthält tierische Erzeugnisse)',
+               'V': 'vegan',
+               'Z': 'Bitte beachten Sie die Auszeichnung am Produkt',
+               '1': 'Mit Konservierungsmittel',
+               '2': 'mit Antioxidationsmittel',
+               '3': 'mit Farbstoff',
+               '4': 'mit Geschmacksverstärker',
+               '5': 'mit Schwefel',
+               '6': 'mit Phosphat',
+               '7': 'geschwärzt',
+               '8': 'gewachst',
+               '9': 'mit Süßungsmittel',
+               '10': 'enthält eine Phenylalaninquelle',
+               '11': 'mit Säurungsmittel',
+               '12': 'mit Stabilisatoren',
+               '13': 'mit Phosphorsäure',
+               '14': 'mit Nitritpökelsalz',
+               '15': 'mit Milcheiweiß',
+               '16': 'koffeinhaltig',
+               '17': 'chininhaltig',
+               '18': 'enthält Schwefeldioxid und Sulfite',
+               '19': 'Milch und Milcherzeugnisse / enthält Laktose',
+               '20': 'Glutenhaltiges Getreide sowie daraus hergestellte Erzeugnisse',
+               '21': 'Soja und Sojaerzeugnisse',
+               '22': 'Sellerie und Sellerieerzeugnisse',
+               '23': 'Senf und Senferzeugnisse',
+               '24': 'Sesamsamen und Sesamerzeugnisse',
+               '25': 'Lupine und Lupinenerzeugnisse',
+               '26': 'Erdnüsse und Erdnusserzeugnisse',
+               '27': 'Fisch und Fischerzeugnisse',
+               '28': 'Krebstiere und Krebstiererzeugnisse',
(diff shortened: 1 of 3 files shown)


duesseldorf.py
```diff
@@ -0,0 +1,163 @@
...
+            # parse menu items
+            meal, raw_legend = parse_menu_items(items)
+            # Studentenwerk Düsseldorf does not maintain a easily readable list of additives etc.
+            # need to do our own parsing against a static list
+            legend = parse_legend(raw_legend)
+            # get prices for each role
+            prices = parse_prices(counter.find('ul', 'price'))
+            # add meal to list
+            canteen.addMeal(date, counter_name, meal, notes=legend, prices=prices)
+        i += 1
```
klemens: This manual counting seems a little brittle. Why don't you iterate over the `counter` divs directly (and merge with the previous function)?

duesseldorf.py
```diff
@@ -0,0 +1,163 @@
+import re
+from urllib.request import urlopen
+
+from bs4 import BeautifulSoup as parse
+
+from pyopenmensa.feed import OpenMensaCanteen
+from utils import Parser
+
+# this will match notes, e.g. "[R]" or "(1,10,12)"
+legend_regex = re.compile('(?:[\[(]{1}[ARFSVZ\d\,]+[\])])')
```
klemens: What is the purpose of the non-capturing group? `{1}` also seems redundant as it's the default anyway.

duesseldorf.py
```diff
@@ -0,0 +1,163 @@
...
+                text = text.replace(li, '').strip()
+                # in addition, remove brackets left over after note deletion
+                for ch in ['(', ')', '[', ']']:
+                    li = li.replace(ch, '')
+                # split notes in list and append master list
+                li = li.split(',')
+                raw_legend.append(li)
+            strings.append(text)
+    # join all strings and return result + raw legend
+    return (", ".join(strings), raw_legend)
```
klemens: Should we remove the ` - Preis ab:` and ` - Preis je 100g:` suffixes from salads and desserts? They seem redundant, as the meals also additionally contain a "Bitte beachten Sie die Auszeichnung am Produkt" note.

duesseldorf.py
Athemis: The problem here is that the counter's names (e.g. "Essen I", "Grill", etc.) are not part of that div, but are provided as separate h2 outside the counter divs. Hence I get two individual lists with no obvious way (at least not obvious to me, hints appreciated ;-) ) to link them to each other directly.

For reference, you may have a look at the [html source of the site itself](http://www.stw-d.de/gastronomie/speiseplaene/mensa-universitaetsstrasse-duesseldorf/).

duesseldorf.py
Athemis: Totally agreed and changed.

duesseldorf.py
```diff
@@ -0,0 +1,163 @@
...
+            legend_tmp = legend_regex.findall(text)
+            for li in legend_tmp:
+                # remove notes from text to obtain a clean string
+                text = text.replace(li, '').strip()
+                # in addition, remove brackets left over after note deletion
+                for ch in ['(', ')', '[', ']']:
+                    li = li.replace(ch, '')
+                # split notes in list and append master list
+                li = li.split(',')
+                raw_legend.append(li)
```
Athemis: Right. Completely overlooked the possibility of duplicate entries here. Thanks!

duesseldorf.py
Athemis: Agreed for "- Preis ab:". As for "- Preis je 100g:" I'm not sure if it's not better to keep this one as the shown prices are really only per 100 g of salad (you'll have to weight it before checkout)

duesseldorf.py
```diff
@@ -0,0 +1,163 @@
...
+            canteen.addMeal(date, counter_name, meal, notes=legend, prices=prices)
+        i += 1
+
+
+def parse_legend(raw_legend):
+    verbose_legend = []  # we'll add parsed legend entries here
+    for li in raw_legend:  # raw_legends contains a two dimensional array
+        for li2 in li:  # second dimension
+            for key in legend_dict:  # iterate over keys and check against legend item (li2)
+                if li2 == key:
```
Athemis: Changed

duesseldorf.py
klemens: I meant the div with the `counter` class instead fo the `counter-table` class:
```html
<div class="counter">
    <h2>Essen I</h2>
    <div class="counter-table">...</div>
</div>
```

Athemis: Oops :-P ok, that was probably too obvious. Changed.

duesseldorf.py
```diff
@@ -0,0 +1,163 @@
...
+               }
+
+# strings to be removed from meal title
+remove_strings = ['- Preis je 100g:',
+                  '- Preis ab:']
+
+
+def parse_url(url, today=False):
+    canteen = OpenMensaCanteen()
+    document = parse(urlopen(url).read())
```
klemens: Looking at the commits by @mswart from today (0e39dc58a4a1d4cc9267f277eea225f1e3ef50d3), we should probably explicitly specify the parser here.

Athemis: Done

klemens:
I rebased, squashed and merged your changes. Thanks again!

I will now try to add the parser to the openmensa.org page.

klemens:
All canteens are online now. I took the addresses from this page: http://www.stw-d.de/gastronomie/oeffnungszeiten/

PR #46, opened 2017-01-17
fix recent changes to erl_nuernberg mensa

theScrabi:
They changed things again. So I had to change things to :/


theScrabi:
halle seams to fail again :/

PR #48, opened 2017-03-23
darmstadt: Hardcode legend data, because it's removed from the HTML

azrdev:


darmstadt.py
```diff
@@ -15,13 +15,35 @@
 def parse_week(url, canteen):
     soup = BeautifulSoup(urlopen(url).read(), 'html.parser')
 
-    try:
-        for legend_tag in soup.select('section.fmc-info p'):
-            canteen.setLegendData(text=' '.join(legend_tag.strings),
+    # TODO: vegetarisch, vegan, Schwein, ... -- indicated via inline SVGs only
+    # legend is a .jpg nowadays, fill it manually
```
klemens: Legends as jpg's are really unfortunate. 😟

PR #54, opened 2017-05-28
add parser for Universität Rostock

wolfbeisz:


rostock.py
```diff
@@ -0,0 +1,93 @@
...
+
+# returns datetime.date object
+def parseDate(str):
+  match = datePattern.match(str)
+  dateStr = match.group(1)
+  dt = datetime.strptime(dateStr, "%d.%m.%Y")
+  #return dt
+  return date(dt.year, dt.month, dt.day)
+
+#def parseNotesOfMeal(td) # name = ""; for span in td.find("span"): text = firstChild; name += text; notes = span.select("small > sup"); format ist [kürzel,kürzel]
```
klemens: It would be very nice to include the notes (ingredients and allergens). The abbreviations are all wrapped in `<strong>` and are either a number or a two-letter abbreviation.

rostock.py
```diff
@@ -0,0 +1,93 @@
...
+# returns integer; in eurocents
+def parsePrice(priceAsString):
```
klemens: Same as with the dates: There is already `convertPrice` which is called automatically when using the `LazyBuilder`, so you can pass in the strings directly.

rostock.py
```diff
@@ -0,0 +1,93 @@
...
+  match = pricePattern.match(priceAsString)
+  return int(match.group(1)) * 100 + int(match.group(2))
+
+def parsePlan(url, internalMensaId, today):
+  #print(internalMensaId)
```
klemens: Please remove these commented debug prints or replace them with actual  comments.

rostock.py
```diff
@@ -0,0 +1,93 @@
...
+  canteen = BaseBuilder()
+  dom = BeautifulSoup(urllib.request.urlopen(url).read(), 'lxml')
+  end = False
+  while (not end):
+    date = parseDate(dom.select('#mensa_date > p')[0].contents[0])
+    #print(date)
+    plan = {date : []}
```
klemens: It seems this isn't used anywhere else?

rostock.py
```diff
@@ -0,0 +1,93 @@
...
+#def parseNotesOfMeal(td) # name = ""; for span in td.find("span"): text = firstChild; name += text; notes = span.select("small > sup"); format ist [kürzel,kürzel]
+# returns integer; in eurocents
+def parsePrice(priceAsString):
+  match = pricePattern.match(priceAsString)
+  return int(match.group(1)) * 100 + int(match.group(2))
+
+def parsePlan(url, internalMensaId, today):
+  #print(internalMensaId)
+  canteen = BaseBuilder()
+  dom = BeautifulSoup(urllib.request.urlopen(url).read(), 'lxml')
```
klemens: If you move this into the loop, you can get rid of the second call to BS below and only need to set the `url` for the next day.

rostock.py
```diff
@@ -0,0 +1,93 @@
+from bs4 import BeautifulSoup
```
klemens: This applies to the whole file: we use 4 spaces instead of 2 for indentation.

klemens:
Thank you for the updates and sorry for the delay. I squashed your commits, added one more small fix: `campus-mensa-wismar` instead of `Campus-mensa-wismar` and merged it to master! 🎉

Do you have an account at http://openmensa.org/? Then you can add this new parser yourself and will get email notifications if something fails. If you don't want to do it, I can also add the parser using my account.

PR #56, opened 2017-10-14
Berlin

azrdev:
maybe @kaifabian wants to tell how/where this shall be hosted in the future

### DIFF
--- a/berlin.py
+++ b/berlin.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+
+import datetime
+import re
+from urllib.request import urlopen, Request
+from urllib.parse import urlencode
+from pyopenmensa.feed import LazyBuilder
+from bs4 import BeautifulSoup
+from utils import Parser
+
+
+def parse_week(url, params, canteen):
+    request = Request(url, data=urlencode(params).encode(), headers={
+        'User-Agent': 'OpenMensa'})  # Python-urllib is blocked
+    response = urlopen(request)
+    soup = BeautifulSoup(response.read(), 'html.parser')
+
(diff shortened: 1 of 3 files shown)


kaifabian:
@azrdev This looks good. Thanks for your efforts!
I will discuss hosting of the new parser with @mswart and get back to you. :)

escrl:
hi I also had a look into it, see my separate pull request. I don't care which one you choose, but this one has to little flaws:
(1) imho if only one price is provided on the website, this should be assumed to be the price for all groups

lenucksi:
Any news on the state of this one?

kaifabian:
IMHO the only open question is whether to use @escrl's or @azrdev's parser. Once that's settled we just need the parser endpoint(s) and user ID of the parser owner. Then @mswart or @jgraichen could replace those informations in the database (so that canteen IDs do not get changed).

escrl:
I already send you details on my parses to info@openmensa.org. As I have no suitable hosting site atm, I run my parser as a cron job on local machines and push result to github at https://github.com/escrl/openmensa-feed-berlin. Only problem that remains with my parser is that automatic updates within openmensa of the feeds don't work, even though I added schedule tags to metadata (e.g. https://raw.githubusercontent.com/escrl/openmensa-feed-berlin/master/mensa_hu_nord/metadata.xml).

mswart:
I try to find time to deal with Berlin this week end.

mswart:
@Patte1808 Sorry for the delay. Over the last weeks both @escrl and myself didn't get any new errror reports. We didn't examined the situation in details and simply assumed everything is working again.
A couple of bugs has been fixed yesterday. All canteens should now have valid meal information.

Please report back if you encounter further issues.

PR #57, opened 2017-10-17
berlin: add parser for university cafeterias

escrl:
Provide meals with annotations by scraping them from Studierendenwerk's
website. If one cafeteria provides only meals without a price tag on a
certain day, it will be assumed to be closed on that day.
Metadata could also be fetched, but this is not done.

### DIFF
--- a/berlin.py
+++ b/berlin.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+
+from pyopenmensa.feed import BaseBuilder, buildPrices
+from bs4 import BeautifulSoup
+from utils import Parser
+import datetime
+import urllib
+
+roles = ('student', 'employee', 'other')
+img_map = {
+    'ampel_gruen_70x65.png': 'Grün (Ampel)',
+    'ampel_gelb_70x65.png': 'Gelb (Ampel)',
+    'ampel_rot_70x65.png': 'Rot (Ampel)',
+    '15.png': 'vegan',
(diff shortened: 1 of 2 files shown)


azrdev:
@escrl do you use the berlin data? I don't, so you might be in a better place to maintain this parser. Do you have any possibility to host it yourself?

escrl:
I plan on using your app from time to time and will encourage some friends to do so, too. 
I arranged some kind of hosting and I am willing to maintain the parser atm. What would I have to do to become maintainer?

escrl:
@mswart I tried to add my parser to openmensa.org myself, but failed. I didn't quite get the idea of the index url. What format is expected json or xml? Do I have to add every single canteen, by hand or should the come magically from the index file?
+@kaifabian what's about the canteens already in the system? I seam not to be able to edit those.

kaifabian:
@escrl yeah, that's correct. and right now, openmensa does not support changing parser ownership. The "officially supported" method would be to remove the old parser and re-create the new canteens... but since this changes canteen ids (and thus would break existing favourites, bookmarks, etc.), @mswart and I thought about editing parser ownership in the database. Once the parser is up and running somewhere, just tell us your user id and the parser URL and we should be able to take care of the rest.

escrl:
just in case you didn't get my mail already I sent one with requested details at info@openmensa . All canteens are now automatically updated at my feed including metadata, making this pull request irrelevant.

PR #58, opened 2017-11-16
Add requirements to setup.py

Y0hy0h:


Y0hy0h:
The build seems to fail on [`wget --output-document=/dev/null --input-file=maybe-urls.txt || true`](https://travis-ci.org/mswart/openmensa-parsers/builds/302782025#L1203). Unfortunately, I do not understand what this means and cannot fix it myself.

Y0hy0h:
Interestingly, #59 which is based on this passed. 🤷‍♂️

Y0hy0h:
Since #59 is based on this and there was no comment against this anywhere, I assume that this change is ok. I propose that we close this PR and simply merge #59, which contains this change as well.

(By the way, sorry that this change is in two PRs... Have learned how to fix this kind of issue, as demonstrated in #63.)

PR #60, opened 2017-11-27
Addapted to changes in the Studierendendenwerk Aachen website

Quarkmus:
Now showing nutrition info

PR #59, opened 2017-11-16
Fix parser for aachen

Y0hy0h:
The menu structure changed for the canteens in Aachen.

Before: https://web.archive.org/web/20170807225026/http://www.studierendenwerk-aachen.de/speiseplaene/academica-w.html  
After: http://www.studierendenwerk-aachen.de/speiseplaene/academica-w.html

This broke the parser, which is sad. ;)

PR #62, opened 2017-12-02
Fixed Aachen Parser

mathismueller:
Hi, 

this should fix the recent problems with the parser for Aachen. I know there is an open pull request to fix mostly the same issues, but it seems broken. This one should actually do the fix and also prevent missing meal names from crashing the parser.

I haven't contributed to too many projects, so I don't know if this is the right way to do things. Let me know if I messed up anything.

### DIFF
--- a/aachen.py
+++ b/aachen.py
@@ -31,12 +31,13 @@ def add_meals_from_table(canteen, table, day):
                 elif namePart.name != 'span':
                     name += namePart.string
         name = name.strip()
-        notes = [legend.get(n, n) for n in notes if n]
-        price_tag = item.find('span', attrs={'class': 'menue-price'})
-        if not price_tag:
-            canteen.addMeal(day, category, name, notes)
-        else:
-            canteen.addMeal(day, category, name, notes, price_tag.text.strip())
+        if name:
+            notes = [legend.get(n, n) for n in notes if n]
+            price_tag = item.find('span', attrs={'class': 'menue-price'})
+            if not price_tag:
+                canteen.addMeal(day, category, name, notes)
+            else:
+                canteen.addMeal(day, category, name, notes, price_tag.text.strip())
 
 
 def parse_day(canteen, day, data):
--- a/aachen.py
+++ b/aachen.py
@@ -19,13 +19,16 @@ def add_meals_from_table(canteen, table, day):
         descs = item.find('span', attrs={'class': 'menue-desc'})
         if not descs:
             return
-        for namePart in descs.children:
-            if type(namePart) is NavigableString:
+        expand = descs.find('span', attrs={'class': 'expand-nutr'})
+        if not expand:
+            return
+        for namePart in expand.children:
+            if type(namePart) is NavigableString and namePart.string is not '+':
                 name += namePart.string
             elif type(namePart) is Tag:
                 if namePart.name == 'sup':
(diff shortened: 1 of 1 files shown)


aachen.py
```diff
@@ -63,11 +67,12 @@ def parse_url(url, today=False):
     regex = '\((?P<name>[\dA-Z]+)\)\s*(?P<value>[\w\s]+)'
     legend = buildLegend(legend, document.find(id='additives').text, regex=regex)
 
-    days = ('montag', 'dienstag', 'mittwoch', 'donnerstag', 'freitag',
-            'montagNaechste', 'dienstagNaechste', 'mittwochNaechste', 'donnerstagNaechste', 'freitagNaechste')
-    for day in days:
-        data = document.find('div', id=day)
-        headline = document.find('a', attrs={'data-anchor': '#' + day})
+    days = ('Montag', 'Dienstag', 'Mittwoch', 'Donnerstag', 'Freitag',
+               'Montag', 'Dienstag', 'Mittwoch', 'Donnerstag', 'Freitag')
```
Y0hy0h: You actually have the exact same issue as the previous PR: https://github.com/mswart/openmensa-parsers/pull/60#discussion_r153779231

Y0hy0h:
There are two PRs already: #59, #60. ;)  
Did you have a look at both of them?

aachen.py
vonMathis: This is not an issue at all. This is more of a design choice. I can change this if you want to, but it works this way. If you look at my other changes, a suffix is appended for the second half of the days in the array.

vonMathis:
I did havea look at the other PRs. However, they don’t seem to handle missing dishes yet. This PR is a quick fix for the most recent changes.

aachen.py
Y0hy0h: Fair enough. However I would argue that suffixing them afterwards is really confusing. In the `days` tuple, it would be very obvious and simple. Maybe you could make this small change? :)

Y0hy0h:
I addressed the issue of empty dishes two days before your PR: https://github.com/mswart/openmensa-parsers/pull/59#issuecomment-348137614  
However, I admit that there was a bug and that it was easy to miss this one commit.

Nonetheless, opening a duplicate PR creates overhead. I would suggest that next time you try commenting and fixing the code on an existing PR. ;)

vonMathis:
Thanks for the comment. I totaly agree with you and will look into how I can fetch other PRs for testing.

It would be great, if any PR that fixes the mentioned problems could be merged ASAP. Waiting for the website to change even more or stabilize eventually is a lost cause in my opinion.

Y0hy0h:
@vonMathis I just googled the issue and found this: https://stackoverflow.com/a/40087766/3287963

Basically, only those who have push rights to OpenMensa can have direct access to commit to a PR. What remains for normal contributors is to submit a "PR on the PR"; if I'd want to fix something in your PR, I would need to go to your repo, open a PR there and have you merge the PR into your branch, so that it shows up here.

mswart:
@mathismueller Thanks for the contribution. But I decided to merge the earliest MR for this parser and close this one for now.
Feel free to update and reopen your MR if you see any remaining issues.

As mentioned in #59 - I would love to delegate the maintenance of this parser to someone else as I do not use any canteen in Aachen. Simply drop a comment if you are interested.

PR #66, opened 2018-01-08
Improve Project's Layout

Y0hy0h:
I had trouble finding the files I wanted, since the root is currently pretty crowded with very different files. Therefore I moved all parsers into a module to make them very easy to find.

Additionally, for the PyOpenMensa dependency, I switched from the git submodule to the pip package and added it to the setup.py. This seems more convenient and is consistent with the other dependencies.

I moved the files used during the travis build into scripts/, since that seems to be the place where all files related to the build reside.

I also refactored config.py to make the side effects a bit more explicit.


Y0hy0h:
I could not figure out why the build fails.

I excluded that moving PyOpenMensa to pip had something to do with it, since a build that reverted that commit failed as well.

Could you point me in a direction that I could look into?

klemens:
The first build (22fb219) failed because `pyopenmensa` could not be found:
```
Traceback (most recent call last):
  File "/usr/share/openmensa-parsers/wsgihandler.py", line 5, in <module>
    from parse import parse
  File "/usr/share/openmensa-parsers/parse.py", line 4, in <module>
    from config import parsers
  File "/usr/share/openmensa-parsers/config.py", line 3, in <module>
    from utils import CanteenPrefixer, ParserRenamer
  File "/usr/share/openmensa-parsers/utils.py", line 8, in <module>
    from pyopenmensa.feed import LazyBuilder
ImportError: No module named 'pyopenmensa'
```
I guess there is nothing installing it.

The second build (14c611b) failed because `parsers` could not be found:
```
Traceback (most recent call last):
  File "/usr/share/openmensa-parsers/wsgihandler.py", line 5, in <module>
    from parse import parse
  File "/usr/share/openmensa-parsers/parse.py", line 4, in <module>
    from config import parsers
  File "/usr/share/openmensa-parsers/config.py", line 48, in <module>
    parsers = register_all_parsers(cities)
  File "/usr/share/openmensa-parsers/config.py", line 35, in register_all_parsers
    register_parser(importlib.import_module('parsers.' + module).parser)
  File "/usr/lib/python3.4/importlib/__init__.py", line 109, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ImportError: No module named 'parsers'
```

Y0hy0h:
Thanks! I will dig deeper into this! :)

Y0hy0h:
Unfortunately, the build process is absolutely confusing me and I cannot figure out what's going on.

I did some experimentation by reverting, commiting and building, but I had no luck in figuring out what the underlying issues are.  
I decided to nuke those experimental commits and leave this PR on the last commit that I initially thought might work.

If you could explain what exactly the issue is, I might try again to fix it, but at the moment I feel quite overwhelmed. ;)

Y0hy0h:
I think it might be an interesting task to refactor the build. Could give me a rundown of what the high level jobs are it accomplishes, so that I could try to recreate them and familiarize myself with it?

mswart:
First of all, thanks for the restructuring. Which the current number of parsers a separation is very helpful.
If the test urls are move to `scripts`, it might be good to rename the `scripts` folder to `test`, `travis` or something like this: the test URLs define which parsers are tested. (_On the other hand it might be reasonable to generated the content based on configuration information from the config and parsers itself.)

The build systems only requires changes very few months to years. So it is not that important that you fully understand how it works in details. Often changes might also require changes within the production system. But I grateful for your motivation to understand and fix it. I hope my following comments help your understanding.

The goal of the jobs is to build and test a Debian packages that can later on be installed. In more detail:
1. It deactivates the pip environment from travis (`before_install`): we want to interact with system tools; the travis own python installations do not help
2. Install all needed python dependencies via system packages (`install` / `scripts/travis-install-debs.sh`)
3. Prepare the environment for package building (`before_script` / `scripts/travis-prepare-packaging.sh`): It generates a new changelog entry in debian/changelog based on the current build id: this way the packages get increasing version numbers and are updates for earlier built packages.
4. The test itself combines building, installation and testing (`script`):
    1. Invokes Debian toolchain to build package (`dpkg-buildpackage -uc -us`): the packages it based upon the files from the debian directory.
    2. Displays the content of the built package: metadata with `dpkg --info` and included files with `dpkg --contents`
   3. install the built package and its required dependencies (`scripts/travis-install-and-setup-deb.sh`): this setups a web-server (`uwsgi`) like the production installation
    4. Call various parsers to check whether they are working: some kind of integration tests.
5. The log file of `uwsgi` is displayed to display occurred exeception during test URL calls (`after_script`)
6. On a push on master the built package it uploaded to be deployed upon the production systems (`after_success`).

I assume the only remaining build failure is the missing pyopenmensa.
The generated `openmensa-parsers` package does only depend upon system packages. Pip dependencies are currently not supported. This is the reason the `setup.py` did not list any dependencies (they were not relevant to my use cases) and one reason why pyopenmensa was an submodule.

Building an own Debian package for pyopenmensa requires work and is error-prone. Travis should therefore receive and include its source code (`pip`, manual `git clone` or an submodule as it was).

Currently I tend to stick with a submodule: it allows to select a specific revision not necessarily an release (there was a time we experimented with an unreleased version of pyopenmensa). For travis and maybe some development systems it would be the only dependencies that would require pip.

Want is your motivation for switching from the submodule to pip?

Y0hy0h:
Thanks for this awesome answer! I understand much better what the goals are, even though I'm a bit lost with how it works exactly. But as you suggested, it might not be worthwhile to fully understand the build process for me. :)

My motivation for switching to pip was just that it's the only method I knew. I thought that it was the official and preferred way to handle dependencies.  
But I understand now why you find it easier to have PyOpenMensa as a submodule. And you also seem to prefer installing debian packages instead of PyPI ones, which is fine with me, since it already runs on my machine. :D

That being said, if you would like me to have a look at something, just tell me. Otherwise I'll simply move on to other stuff. :)

Y0hy0h:
Ok, I'm now completely lost on what's going on. I've reverted the changes made, but the build keeps not finding the scripts. 💁

klemens:
It seems the problem is that the scripts are no longer executable (mode 644 instead of 755).

Y0hy0h:
Finally it's not my code anymore that crashes the build! :)

I'll wait until someone approves #67 and it is merged, then I'll merge the fix into this PR and it should hopefully, finally pass.

azrdev:
@Y0hy0h not wanting to derail or demotivate you, but you might also look for parsers yet to be written, or give the android app some love (it's abandoned for quite some time now, afaik) :)

Y0hy0h:
@azrdev, yes, I'm currently mostly refactoring, in order to make it easier for us (especially for me) to work on this project.  
I currently plan on taking a look at PyOpenMensa and refactoring (or creating a new version) that fixes some issues I had with it in the Aachener parser.  
For the Android app, I'm hesitant, because I'm tinkering with Flutter and really loving that, so I don't really want to get into native Android...

To sum up, I'm cleaning up some bits and pieces that I'm finding, and after that it should (hopefully) be easier for everyone to contribute. ;)  
However, thanks for your hints! I will keep them in the back of my head.

Y0hy0h:
Finally! :)

Y0hy0h:
A big thanks to @klemens for pointing me in the right direction twice! :D Without you, I wouldn't have managed to figure anything out.

PR #67, opened 2018-01-10
Fix Rostock's URL after apparent change

Y0hy0h:
The old URL, https://www.studentenwerk-rostock.de/de/mensen/speiseplaene.html, redirects to https://www.stw-rw.de/de/mensen/speiseplaene.html, which is empty. Clicking the button to see the canteens opens https://www.stw-rw.de/de/mensen-und-cafeterien/speiseplaene.html.

Setting the parser to this new URL makes it run again and the first entry in the output of Mensa Süd matches the website.

### DIFF
--- a/rostock.py
+++ b/rostock.py
@@ -61,7 +61,7 @@ def parse_url(url, today=False):
     return parsePlan(splitted[0], splitted[1], today)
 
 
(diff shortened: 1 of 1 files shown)


Y0hy0h:
It would be nice if someone could approve the change. Since this would be my very first merge, I would like to have a sanity check. ;)

azrdev:
looks good to me!

PR #71, opened 2018-02-08
OpenMensa model

Y0hy0h:
The major change concerns the addition of new model classes. The are found in the root in the `openmensa_model` module. They contain classes for a `Canteen`, `Day` (and `DayClosed`), `Category`, `Meal`, `PriceWithRoles` and `Role`, which each handle their own XML generation.

This allowed me to break down the Aachener parser into more concise functions, that each return a model object for e. g. a `Meal` and then aggregate them into a `Category`, then a `Day` etc.

I additionally created a model just for the Aachener parser, because the Aachener menu differs slightly form the OpenMensa model. E. g., the price is fixed for a category, not per meal. But it is straight forward to convert this Aachener model into the OpenMensa variant in the end.

I believe that this approach to generating the XML is easier to maintain than the centralized XML generator in PyOpenMensa's `BaseBuilder`, since it breaks down the XML into the data classes. It also helps structure the parsers' code, since it doesn't require them to input all the meals information in one single line, but encourages splitting up the parsing to the individual components.


mswart:
What is the difference between `openmensa_model` and `parser/canteen`? It do not really understand their individual propose.
But as mentioned in the old #70, please add such generic feed generation helps to pyopenmensa. In addition, I think such abstract helper class should have unit tests.

Y0hy0h:
I added validation of the generated XML against the schema. This was also proposed in #40.

It was discussed, whether the schema .xsd should live in this repository. I think it should be already downloaded to prevent fetching it constantly over the network. However, fundamentally this questions ties back to one we had previously: Whether this model should be in this repository or in PyOpenMensa. I think, for now we should keep it here until it is mature enough to be released as part of PyOpenMensa.

Y0hy0h:
I've added helper classes that will allow me to more easily refactor the Aachener parser. They all produce a OpenMensa model object and they can be used together.

I believe that the builder paradigm is a very powerful approach to help build parsers more easily, while allowing each parser to choose what to do itself and what to let the helpers do.

What do you think?

Y0hy0h:
I have used the model to refactor Aachen in #73. I believe that this model is usable now (though surely not perfect). We could therefore move it to PyOpenMensa.

klemens:
These are still both `openmensa_model` and `parser/canteen`. Did you intend to delete the latter?

Y0hy0h:
Yes, I was totally under the impression I had already deleted `parsers/canteen`. Fixed that.

Y0hy0h:
I've opened a PR on PyOpenMensa: https://github.com/mswart/pyopenmensa/pull/12

During integration in PyOpenMensa, I chose to shorten the name from `openmensa_model` to `model`, allow imports directly from `pyopenmensa.model` and rename the `helpers` to `builders`.

Y0hy0h:
Reopened, because it seems unlikely that it will be merged into PyOpenMensa (https://github.com/mswart/pyopenmensa/pull/12).

Y0hy0h:
@klemens I would like to close this PR. Would you be up to either merging it or closing it?

If there are any fixes you would like, I could work on them, but I don't want to spend significant time on this. Therefore, it's completely ok for you to decide you do not want to maintain this and close it. ;)

Y0hy0h:
@klemens Any decisions? Otherwise I will just close this PR and not merge the changes here.

klemens:
@Y0hy0h: Sorry for not answering earlier.

I haven't looked at the code for a while, but here are some random thoughts: If we do this, we should do it in PyOpenMensa and not as a parallel implementation, but replacing the existing implementation, without changing the existing API (much). This is because the existing API works fine for most parsers IMO (and we don't want to break compatibility). Without support from Malte (which seems unlikely), we have to fork PyOpenMensa, as I don't have access to either the git repo or the pypi package.

Now all this is a lot of work and I currently don't have a lot of free time, which is why there hasn't been any activity here. 😕

PR #73, opened 2018-03-03
Aachen refactor

Y0hy0h:
Refactor Aachener parser using OpenMensa model from #71.

It is based on both the snapshot tests (#96) and the model (#71) branches. Both of these should be merged before this PR.

## Fixes

- Only some meals were subsidized and had different prices for students vs others, but the old parser had the two prices for all meals.
- Closed days and closed (unavailable) meals were not recognized.


Y0hy0h:
Closed because #71 was closed. I will fix the issues that this PR addresses in a separate rewrite.

PR #74, opened 2018-03-25
Add parser for mensaar.de api

### LINKED ISSUE #1: Support Saarbrücken

ialokim:
This PR adds support for the API provided by [mensaar.de](https://mensaar.de/) and thus for the following universities:
* Mensa / Mensacafe Saarbrücken
* Mensa Homburg
* Cafeteria HTW Göttelborn

Fixes #1

### DIFF
--- a/parsers/saarland.py
+++ b/parsers/saarland.py
@@ -2,6 +2,7 @@ from urllib.request import *
 
 import json
 import datetime
+import sys
 
 from pyopenmensa.feed import LazyBuilder
 
@@ -53,9 +54,7 @@ NOTICES_PREFIX_ALLERGENS = "enthält"
 NOTICES_PREFIX_OTHERS = "mit"
 
 
-def get_notices(notices, sub_notices=False):
-    global base_data
-
+def get_notices(base_data, notices, sub_notices=False):
     notices_special = []
     notices_allergens = []
     notices_others = []
@@ -99,15 +98,15 @@ def build_subnotices(notices, prefix):
     return notices
 
 
-def build_notes(notices, components):
+def build_notes(base_data, notices, components):
     components_all = []
 
-    notices_list = get_notices(notices)
+    notices_list = get_notices(base_data, notices)
 
     for component in components:
         component_string = '%s %s' % (NOTICES_PREFIX_COMPLEMENT, component['name'])
         if len(component['notices']) > 0:
-            component_notices_list = get_notices(component['notices'], True)
+            component_notices_list = get_notices(base_data, component['notices'], True)
             component_string += ' (%s)' % ', '.join(component_notices_list)
         components_all.append(component_string)
 
@@ -134,9 +133,7 @@ def build_location(description):
 
 
 def parse_url(url, today=False):
-    global base_data
-
-    load_base_data()
+    base_data = load_base_data()
 
     canteen = LazyBuilder()
     with urlopen(url) as response:
@@ -157,26 +154,27 @@ def parse_url(url, today=False):
                 if 'knownMealId' in meal:
                     # This is meant to allow recognizing recurring meals,
                     # for features like marking meals as favorites.
-                    # Up to now, not really used in the mensaar.de API.
-                    print('knownMealId: %s' % meal['knownMealId'])
+                    # Up to now, not really used in the mensaar.de API,
+                    # nor functional in this API parser.
+                    # The meal will still be recognized as every other meal.
+                    print('knownMealId: %s' % meal['knownMealId'], file=sys.stderr)
 
                 meal_name = meal['name']
                 if 'category' in meal:
                     meal_name = '%s: %s' % (meal['category'], meal_name)
 
                 meal_notes = (
+                    # The description is typically the location
+                    # (but not required to be by the API specification).
                     build_location(counter_description) +
                     build_hours(counter_hours) +
-                    build_notes(meal['notices'], meal['components']))
+                    build_notes(base_data, meal['notices'], meal['components']))
 
                 meal_prices = {}
                 if 'prices' in meal:
                     prices = meal['prices']
                     for role in prices:
-                        if role in base_data['roles']:
-                            meal_prices[base_data['roles'][role]] = prices[role]
-                        else:
-                            print('Ignoring price for unknown role %s.' % role)
+                        meal_prices[base_data['roles'][role]] = prices[role]
 
                 if 'pricingNotice' in meal:
                     meal_notes.append(meal['pricingNotice'])
@@ -191,6 +189,7 @@ def load_base_data():
     with urlopen(URL_BASE + URL_BASE_DATA) as response:
         data = json.loads(response.read().decode())
 
+    base_data = {}
     base_data['roles'] = {}
 
     base_data['notices'] = data['notices']
@@ -204,7 +203,6 @@ def load_base_data():
     for role in data['priceTiers']:
         if role not in ROLES:
             # Found an unknown price tier
-            # All prices for this price tier will be ignored.
             # Please consider updating the parser!
             raise RuntimeError(
                 'Unknown price tier: %s (displayName: %s)' %
@@ -212,8 +210,8 @@ def load_base_data():
         else:
             base_data['roles'][role] = ROLES[role]
 
+    return base_data
 
-base_data = {}
 
 parser = Parser('saarland',
                 handler=parse_url,
--- a/parsers/saarland.py
+++ b/parsers/saarland.py
@@ -24,6 +24,14 @@ URL_MENU = "getMenu/"
 # mensaar.de always returns an UTC date
 UTC_DATE_STRING = "%Y-%m-%dT%H:%M:%S.%fZ"
 
+LOCATIONS = [
+    'sb',
+    'hom',
+    'musiksb',
+    'htwgtb',
+    'mensagarten'
+]
+
 ROLES = {
     's': 'student',
     'g': 'other',
@@ -128,6 +136,8 @@ def build_location(description):
 def parse_url(url, today=False):
     global base_data
 
+    load_base_data()
+
     canteen = LazyBuilder()
     with urlopen(url) as response:
         data = json.loads(response.read().decode())
@@ -177,30 +187,37 @@ def parse_url(url, today=False):
     return canteen.toXMLFeed()
 
 
-with urlopen(URL_BASE + URL_BASE_DATA) as response:
-    data = json.loads(response.read().decode())
+def load_base_data():
+    with urlopen(URL_BASE + URL_BASE_DATA) as response:
+        data = json.loads(response.read().decode())
+
+    base_data['roles'] = {}
+
+    base_data['notices'] = data['notices']
+    for loc in data['locations']:
+        if loc not in LOCATIONS:
+            # Found an unknown location
+            # Please consider updating the parser!
+            raise RuntimeError(
+                'Unknown location: %s (displayName: %s)' %
+                (loc, data['locations'][loc]))
+    for role in data['priceTiers']:
+        if role not in ROLES:
+            # Found an unknown price tier
+            # All prices for this price tier will be ignored.
+            # Please consider updating the parser!
+            raise RuntimeError(
+                'Unknown price tier: %s (displayName: %s)' %
+                (role, data['priceTiers'][role]))
+        else:
+            base_data['roles'][role] = ROLES[role]
+
+
 base_data = {}
-base_data['locations'] = []
-base_data['roles'] = {}
-
-base_data['notices'] = data['notices']
-for loc in data['locations']:
-    base_data['locations'].append(loc)
-for role in data['priceTiers']:
-    if role not in ROLES:
-        # Found an unknown price tier
-        # All prices for this price tier will be ignored.
-        # Please consider updating the parser!
-        raise RuntimeError(
-            'Unknown price tier: %s (displayName: %s)' %
-            (role, data['priceTiers'][role]))
-    else:
-        base_data['roles'][role] = ROLES[role]
-
 
 parser = Parser('saarland',
                 handler=parse_url,
                 shared_prefix=URL_BASE + URL_MENU)
 
-for loc in base_data['locations']:
+for loc in LOCATIONS:
     parser.define(loc, suffix=loc)
--- a/parsers/saarland.py
+++ b/parsers/saarland.py
@@ -1,0 +1,177 @@
+from urllib.request import *
+
+import json
+import datetime
+
+from pyopenmensa.feed import LazyBuilder
+
+from utils import Parser, Source
+
+
+API_VERSION = "1"
+# please don't copy this key, contact felix@fefrei.de to get your own instead
+API_KEY = "A1RT5M56" + "ibsGaGNZBJQg"
+APP_VERSION = "1"
+SUPPORTED_LANGUAGES = ["de", "en"]
+
+URL_BASE = (
+    'https://mensaar.de/api/%s/%s/%s/%s/'
+    % (API_VERSION, API_KEY, APP_VERSION, SUPPORTED_LANGUAGES[0])
+)
+URL_BASE_DATA = "getBaseData"
+URL_MENU = "getMenu/"
+
+UTC_DATE_STRING = "%Y-%m-%dT%H:%M:%S.%fZ"
+
+ROLES = {
+    's': 'student',
+    'g': 'other',
+    'm': 'employee'
+}
+
+SPECIAL_NOTICES = [
+    'su',   # geschwefelt
+    'bl',   # geschwärzt
+    'gw',   # gewachst
+    've',   # vegetarisch
+    'bio',  # biologisches Essen
+    'nla',  # laktosefrei
+    'vn'    # vegan
+]
+
+
+def get_notices(notices):
+    global base_data
+
+    notices_special = []
+    notices_allergens = []
+    notices_others = []
+
+    for notice_short in notices:
+        notice = base_data['notices'][notice_short]
+        if notice['isAllergen']:
+            notices_allergens.append(
+                'enthält %s (Allergen)' % notice['displayName'])
+        else:
+            if notice_short == 'nsf':  # ohne Schweinefleisch
+                notices_others.append(notice['displayName'])
+            elif notice_short in SPECIAL_NOTICES:
+                notices_special.append(notice['displayName'])
+            else:
+                notices_others.append('mit %s' % notice['displayName'])
+
+    return notices_special + notices_allergens + notices_others
+
+
+def build_notes(notices, components):
+    global base_data
+
+    components_all = []
+
+    notices_list = get_notices(notices)
+
+    for component in components:
+        component_string = 'dazu ' + component['name']
+        if len(component['notices']) > 0:
+            component_notices_list = get_notices(component['notices'])
+            component_string += ' (%s)' % ', '.join(component_notices_list)
+        components_all.append(component_string)
+
+    return components_all + notices_list
+
+
+def build_hours(opening_hours):
+    if not opening_hours:
+        return []
+
+    start = datetime.datetime.strptime(
+        opening_hours['start'], UTC_DATE_STRING).time()
+    end = datetime.datetime.strptime(
+        opening_hours['end'], UTC_DATE_STRING).time()
+    return [
+        'verfügbar von %s bis %s Uhr' %
+        (datetime.time.strftime(start, "%H:%M"),
+            datetime.time.strftime(end, "%H:%M"))
+    ]
+
+
+def build_location(description):
+    return ['Ort: %s' % description] if description else []
+
+
+def parse_url(url, today=False):
+    global base_data
+
+    canteen = LazyBuilder()
+    with urlopen(url) as response:
+        data = json.loads(response.read().decode())
+
+    for day in data['days']:
+        date = day.get('date')
+
+        if today and (datetime.datetime.now().date() !=
+                      datetime.datetime.strptime(date, UTC_DATE_STRING).date()):
+            continue
+
+        for counter in day['counters']:
+            counter_name = counter['displayName']
+            counter_description = counter['description']
+            counter_hours = counter.get('openingHours')
+            counter_color = counter.get('color')
+
+            for meal in counter['meals']:
+                if 'knownMealId' in meal:
+                    print('knownMealId: %s' % meal['knownMealId'])
+
+                meal_name = meal['name']
+                if 'category' in meal:
+                    meal_name = '%s: %s' % (meal['category'], meal_name)
+
+                meal_notes = (
+                    build_location(counter_description) +
+                    build_hours(counter_hours) +
+                    build_notes(meal['notices'], meal['components']))
+
+                meal_prices = {}
+                if 'prices' in meal:
+                    prices = meal['prices']
+                    for role in prices:
+                        if role in base_data['roles']:
+                            meal_prices[base_data['roles'][role]] = prices[role]
+                        else:
+                            print('Ignoring price for unknown role %s.' % role)
+
+                if 'pricingNotice' in meal:
+                    meal_notes.append(meal['pricingNotice'])
+
+                canteen.addMeal(date, counter_name,
+                                meal_name, meal_notes, meal_prices)
+
+    return canteen.toXMLFeed()
+
+
+with urlopen(URL_BASE + URL_BASE_DATA) as response:
+    data = json.loads(response.read().decode())
+base_data = {}
+base_data['locations'] = []
+base_data['roles'] = {}
+
+base_data['notices'] = data['notices']
+for loc in data['locations']:
+    base_data['locations'].append(loc)
+for role in data['priceTiers']:
+    if role not in ROLES:
+        print('New unknown price tier: %s (displayName: %s)' %
+              (role, data['priceTiers'][role]))
+        print('All prices for this price tier will be ignored.')
+        print('Please consider updating the parser!')
+    else:
+        base_data['roles'][role] = ROLES[role]
+
+
+parser = Parser('saarland',
+                handler=parse_url,
+                shared_prefix=URL_BASE + URL_MENU)
+
+for loc in base_data['locations']:
+    parser.define(loc, suffix=loc)
--- a/parsers/saarland.py
+++ b/parsers/saarland.py
@@ -21,6 +21,7 @@ URL_BASE = (
 URL_BASE_DATA = "getBaseData"
 URL_MENU = "getMenu/"
 
+# mensaar.de always returns an UTC date
 UTC_DATE_STRING = "%Y-%m-%dT%H:%M:%S.%fZ"
 
 ROLES = {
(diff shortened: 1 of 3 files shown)


ialokim:
Up to now, only support the German language version of the API, I'm not really sure how to expose a translated menu to OpenMensa :thinking:

fefrei:
In practice, most strings aren't actually translated anyway (yet) because of internal difficulties `¯\_(ツ)_/¯`

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+
+UTC_DATE_STRING = "%Y-%m-%dT%H:%M:%S.%fZ"
+
+ROLES = {
+    's': 'student',
+    'g': 'other',
+    'm': 'employee'
+}
+
+SPECIAL_NOTICES = [
```
fefrei: I don't really like that you need to hard-code certain notice keys here. Maybe we can extend the `getBaseData` call to provide you with the categorization we need? What would we need to add, exactly?

ialokim: I'm building strings to use as notes on the meal object (see lines [58](https://github.com/mswart/openmensa-parsers/pull/74/files/8a2c2afd2ba9c57f951bc042f568b9529dc39cda#diff-91dab909c61ef0b60c3823e035632147R58) and [68](https://github.com/mswart/openmensa-parsers/pull/74/files/8a2c2afd2ba9c57f951bc042f568b9529dc39cda#diff-91dab909c61ef0b60c3823e035632147R68)), because of a pretty limited feed specification (see [openmensa#34](/openmensa/openmensa/issues/34)) with no differentiation between allergic, nutritional or ingredients' information.

Regarding the mensaar.de API, the only differentiation between `notices` is provided by the `isAllergen` value, but most of the `notices` are nouns (and could be formatted as in the lines linked to above), but these "special" ones are adjectives that does not fit within the string. I therefore needed to handle them separately to avoid a `note` like `mit geschwefelt`, for example.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+    for notice_short in notices:
+        notice = base_data['notices'][notice_short]
+        if notice['isAllergen']:
+            if sub_notices:
+                notices_allergens.append(notice['displayName'])
+            else:
+                notices_allergens.append(
+                    'enthält %s' % notice['displayName'])
+        else:
+            if notice_short == 'nsf':  # ohne Schweinefleisch
```
ialokim: Another special case is `nsf`, the only value that states explicitly the _absence_ of an ingredient, so here too, I had to avoid a `note` as `mit ohne Schweinefleisch`.

parsers/saarland.py
ialokim: I'm not really sure whether it is necessary to extend the API, but it would be a nice improvement to mark these cases differently.

parsers/saarland.py
fefrei: Indeed, I'd say this is an oversight in the API. It's fixed now:

```
        "nu": {
            "displayName": "Schalenfrüchte (Nüsse)",
            "isAllergen": true,
            "isNegated": false
        },
        "nsf": {
            "displayName": "ohne Schweinefleisch",
            "isAllergen": false,
            "isNegated": true
        },
```

I hope this helps!

parsers/saarland.py
fefrei: Hm, in that case, I think your list is incomplete: “mit kann Spuren von Nüssen beinhalten” isn't right either, but `nu` is not in your exception list.

I'd be happy to extend the API here, but I currently don't see which exact criterion I'd have to set up to help you and others with this problem. Given that English translations aren't currently available anyways, maybe checking whether the notice text starts with a capital letter might be enough? The only exception to this rule seems to be `Aus zum Teil fein zerkleinertem Fleisch`, and I'm removing the capital letter right now 😉

parsers/saarland.py
ialokim: Wow, you're quick! Thanks a lot!

fefrei: That's what happens when the API isn't operated by the Studentenwerk itself, but an enthusiastic student – API changes are just one push away 🎉

parsers/saarland.py
ialokim: > Hm, in that case, I think your list is incomplete: “mit kann Spuren von Nüssen beinhalten” isn't right either, but nu is not in your exception list.

Uuuuh, you're right, I've missed that one! Seems like there are no meals with nuts within the next four days :laughing: 

> Given that English translations aren't currently available anyways, maybe checking whether the notice text starts with a capital letter might be enough?

That would be an alternative, indeed. I'm not sure which one I prefer, though.

> I'd be happy to extend the API here, but I currently don't see which exact criterion I'd have to set up to help you and others with this problem.

Just thinking, if you would introduce another key called `isAdjective` or `isDescriptive` or something similar... or a key for the nouns called `isComponent` or `isIngredient`...?

fefrei: I'm totally open to that, but I haven't found something that sounds totally consistent yet. For example, *kann Spuren von Nüssen beinhalten* is in some sense an ingredient, but for your purpose, you'd clearly want `isIngredient: false`, right?

There are even more corner-cases: What about *Krebstiere, Krusten- und Schalentiere*? These are clearly ingredients, or components, or however you want to call them, but you cannot say *mit Krebstiere, Krusten- und Schalentiere* – you'd need to say *mit Krebstiere**n**, Krusten- und Schalentiere**n***.

Another workaround would be for the API to always return an alternative “descriptive form”, which would never be a single substantive, but always a describing phrase. But that also doesn't feel like a good solution for me.

ialokim: >  For example, kann Spuren von Nüssen beinhalten is in some sense an ingredient, but for your purpose, you'd clearly want isIngredient: false, right?

Is that a possible future example? I didn't get it by the API up to now :thinking: 
I think these kind of cases shouldn't have a `displayName` as _kann Spuren von Nüssen beinhalten_ but only _Nüsse_ and another key, some sort of `isPossibility` or `couldBe`... It would be up to the client to handle this case accordingly.

> What about Krebstiere, Krusten- und Schalentiere?

In this case, I've been lucky because it's marked as `isAllergen` and _enthält Krebstiere, Krusten- und Schalentiere_ works. But that's the problem of the German language and Its accusative forms :see_no_evil: 

> Another workaround would be for the API to always return an alternative “descriptive form”, which would never be a single substantive, but always a describing phrase. But that also doesn't feel like a good solution for me.

Nor to me, to be honest. The best possible solution IMHO would be to restrict the possible values of `displayName` (e.g. to one word) and to abstract other properties to standard keys. But that implies some changes on client-side (e.g. _ohne Schweinefleisch_ should only have _Schweinefleisch_ as `displayName`, while the client has to recognize the `isNegative` key and add an _ohne_ to the displayed form).

Or to keep backward compatibility, you could introduce a new key `name` with the restriction mentioned above and keep the `displayName` key as a sort of "descriptive form" (as it already is in some cases).

fefrei: > Is that a possible future example? I didn't get it by the API up to now 🤔

The `isIngredient` flag was just an example as I'm not sure what the best option is (yet).

> I think these kind of cases shouldn't have a displayName as kann Spuren von Nüssen beinhalten but only Nüsse and another key, some sort of isPossibility or couldBe... It would be up to the client to handle this case accordingly.

Hm, there is some truth to that, but that would be a breaking change (at least semantically), and this text was provided like so from Studentenwerk, so I don't really want to change it.

> In this case, I've been lucky because it's marked as isAllergen and enthält Krebstiere, Krusten- und Schalentiere works. But that's the problem of the German language and Its accusative forms 🙈

You're a lucky guy 😉

>  abstract other properties to standard keys

That sounds good, but it's actually a bit tricky. There are at least two cases (*components* and *properties*, although the separation is sometimes a bit tricky), negation, *and* and *or, ...

ialokim: > The isIngredient flag was just an example as I'm not sure what the best option is (yet).

I was referring to the concrete example of _kann Spuren von Nüssen enthalten_ :smile: 

> Hm, there is some truth to that, but that would be a breaking change (at least semantically), and this text was provided like so from Studentenwerk, so I don't really want to change it.

That's why I thought of leaving this text provided from Studentenwerk as it is for `displayName` and introducing the new (restricted) `name` key.

> There are at least two cases (components and properties, although the separation is sometimes a bit tricky)

In which cases would it be tricky?

> and and *or, ...

I'm not sure what do you mean by _and_ and _or_?

fefrei: > I was referring to the concrete example of kann Spuren von Nüssen enthalten 😄

Oops. Turns out the configuration file provided by Studentenwerk contains both

    "nu": {
        "displayName": {"de": "Schalenfrüchte (Nüsse)", "en": "Nuts"},
        "isAllergen": true
    },

and

    "nu": {
        "displayName": {"de": "kann Spuren von Nüssen beinhalten", "en": "may contain traces of nuts"},
        "isAllergen": false
    },

so we actually only see the former.

> In which cases would it be tricky?

How would you classify *aus zum Teil fein zerkleinertem Fleisch*? In some sense, it's both component and property.

> I'm not sure what do you mean by and and or?

I'm talking about weird notices like *Macadamia-oder-Queenslandnüsse* or *Krebstiere, Krusten- und Schalentiere*. None of these are made up, unfortunately 😟

ialokim: > How would you classify aus zum Teil fein zerkleinertem Fleisch? In some sense, it's both component and property.

Okay, I see. So what do you think about introducing the three different notices groups _components_, _properties_ and _other_, while _other_ would cover these (very few) special cases. So, every _notice_ object would have a key `group` or (`type`) with the only accepted values `component` (or `ingredient`), `property` and `other` (or simply `note`).

> I'm talking about weird notices like Macadamia-oder-Queenslandnüsse or Krebstiere, Krusten- und Schalentiere.

Hmmmm, then why not better leaving these notices' names together? If they are only composed of nouns and _und_/_oder_ :thinking:

fefrei: Hm, that could work, although `other` still feels like cheating. I've asked the IT department of Studentenwerk for their opinion, maybe they have some comments.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+            notices_others[-2] = (
+                '%s und %s' % (notices_others[-2], notices_others[-1]))
+            del notices_others[-1]
+
+    return (notices_special + notices_allergens +
+            notices_others + notices_others_last)
+
+
+def build_notes(notices, components):
+    global base_data
```
klemens: `base_data` is not used in this function.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+        if len(notices_allergens) > 1:
+            notices_allergens[-2] = (
+                '%s und %s' % (notices_allergens[-2], notices_allergens[-1]))
+            del notices_allergens[-1]
+        if len(notices_others) > 0:
+            notices_others[0] = 'mit %s' % notices_others[0]
+        if len(notices_others) > 1:
+            notices_others[-2] = (
+                '%s und %s' % (notices_others[-2], notices_others[-1]))
+            del notices_others[-1]
```
klemens: As both these do the same thing, you could mode this into a separate function. Also, I am not sure if separating the last two by „und“ is really necessary, especially as single notes can already contain „und“:
> dazu Chili-Kräuterdip (enthält Soja, Milch und Laktose, Senf und Weizen)

parsers/saarland.py
klemens: Could you incorporate this change?

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+def get_notices(notices, sub_notices=False):
+    global base_data
+
+    notices_special = []
+    notices_allergens = []
+    notices_others = []
+    notices_others_last = []
+
+    for notice_short in notices:
+        notice = base_data['notices'][notice_short]
```
klemens: Are all notices always contained in the base data?

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+
+        for counter in day['counters']:
+            counter_name = counter['displayName']
+            counter_description = counter['description']
+            counter_hours = counter.get('openingHours')
+            counter_color = counter.get('color')
+
+            for meal in counter['meals']:
+                if 'knownMealId' in meal:
+                    print('knownMealId: %s' % meal['knownMealId'])
```
klemens: There is currently one known meal in the base data:
```json
"knownMeals": {
  "gebfl": {
    "displayName": "Gebackener Fleischkäse"
  }
}
```

Should this be used somehow? Do these meals maybe have no displayName, which would cause the `counter['displayName']` above to fail?

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+
+base_data['notices'] = data['notices']
+for loc in data['locations']:
+    base_data['locations'].append(loc)
+for role in data['priceTiers']:
+    if role not in ROLES:
+        print('New unknown price tier: %s (displayName: %s)' %
+              (role, data['priceTiers'][role]))
+        print('All prices for this price tier will be ignored.')
+        print('Please consider updating the parser!')
```
klemens: I don't think printing these notes (also above) really helps as probably no one notices these anyway. You could throw an exception, which will get noticed, as openmensa.org sends emails when parsers fail, or simply ignore unknown roles.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+                    meal_notes.append(meal['pricingNotice'])
+
+                canteen.addMeal(date, counter_name,
+                                meal_name, meal_notes, meal_prices)
+
+    return canteen.toXMLFeed()
+
+
+with urlopen(URL_BASE + URL_BASE_DATA) as response:
+    data = json.loads(response.read().decode())
```
klemens: Requesting this in the main scope is not a good idea. It leads to executing network requests when simply including the module, which happens even if the parser is not used.
Also depending on how this is hosted, it might be that this is executed once when starting the application and then never updated again.

So I suggest simply moving this into a separate function that gets called in `parse_url` and define the parsers manually (they currently (see #78) have to be defined manually in openmesa.org anyway and they probably don't change that often). This way you can also avoid the `global` definitions.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+                    print('knownMealId: %s' % meal['knownMealId'])
+
+                meal_name = meal['name']
+                if 'category' in meal:
+                    meal_name = '%s: %s' % (meal['category'], meal_name)
+
+                meal_notes = (
+                    build_location(counter_description) +
+                    build_hours(counter_hours) +
+                    build_notes(meal['notices'], meal['components']))
```
klemens: Just a quick info: While the generated xml keeps the notes in order, openmensa.org doesn't store them in order (see #77 for discussion). This is also something that we might want to change (I haven't really looked into this yet).

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+        if len(component['notices']) > 0:
+            component_notices_list = get_notices(component['notices'], True)
+            component_string += ' (%s)' % ', '.join(component_notices_list)
+        components_all.append(component_string)
+
+    return components_all + notices_list
+
+
+def build_hours(opening_hours):
+    if not opening_hours:
```
klemens: NIT: This should be `if opening_hours is None`.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+SUPPORTED_LANGUAGES = ["de", "en"]
+
+URL_BASE = (
+    'https://mensaar.de/api/%s/%s/%s/%s/'
+    % (API_VERSION, API_KEY, APP_VERSION, SUPPORTED_LANGUAGES[0])
+)
+URL_BASE_DATA = "getBaseData"
+URL_MENU = "getMenu/"
+
+UTC_DATE_STRING = "%Y-%m-%dT%H:%M:%S.%fZ"
```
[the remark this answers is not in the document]
fefrei: Yes. As per the spec, the `date` field on a *day* object contains *a date representing the corresponding day at midnight, UTC*.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+            for meal in counter['meals']:
+                if 'knownMealId' in meal:
+                    print('knownMealId: %s' % meal['knownMealId'])
+
+                meal_name = meal['name']
+                if 'category' in meal:
+                    meal_name = '%s: %s' % (meal['category'], meal_name)
+
+                meal_notes = (
+                    build_location(counter_description) +
```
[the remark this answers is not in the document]
fefrei: Typically yes, but the API doesn't promise that, this might also contain other human-readable information.

parsers/saarland.py
fefrei: Yes, every notice must be specified in the `getBaseData` call. (An exception would be if the base data is stale. This is very unlikely, as new notices are introduced rarely.)

parsers/saarland.py
fefrei: This is meant to allow recognizing recurring meals, for features like marking meals as favorites. (Simply matching the title isn't enough because there often are minor differences.)

So far, I couldn't actually convince Studentenwerk Saarland to build the list of recurring meals, so this list only contains this single entry as an example.

parsers/saarland.py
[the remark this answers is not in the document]
Y0hy0h: Maybe you could add a comment linking to the spec?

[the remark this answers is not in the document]
fefrei: The specification isn't public (so far).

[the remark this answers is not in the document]
Y0hy0h: Oh, ok. Maybe just a comment stating the reliance on the spec? Just so that when it changes and someone reads the code, there is an explicit dependency stated, so that they'll catch it more likely.

Something like `Mensaar.de always returns an UTC date` or similar? Or what do you think of such a comment?

parsers/saarland.py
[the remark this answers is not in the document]
ialokim: Okay, I'm not quite sure how to handle this case of checking whether it is a location or not. The `build_location` function could check for all known locations prior to prepending "Ort: ", but this would be hardcoded information again.

So better drop the `build_location` logic or is there a way changing the API to always get a location with a predefined key?

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if today and (datetime.datetime.now().date() !=
+                      datetime.datetime.strptime(date, UTC_DATE_STRING).date()):
+            continue
+
+        for counter in day['counters']:
+            counter_name = counter['displayName']
+            counter_description = counter['description']
+            counter_hours = counter.get('openingHours')
+            counter_color = counter.get('color')
```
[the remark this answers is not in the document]
ialokim: You're right, I didn't find a way to reflect color information in openmensa feeds. I'll remove it.

parsers/saarland.py
ialokim: You're right, I've removed it.

parsers/saarland.py
ialokim: Changed.

parsers/saarland.py
ialokim: Done.

parsers/saarland.py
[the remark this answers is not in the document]
ialokim: I agree, it's added.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+
+    canteen = LazyBuilder()
+    with urlopen(url) as response:
+        data = json.loads(response.read().decode())
+
+    for day in data['days']:
+        date = day.get('date')
+
+        if today and (datetime.datetime.now().date() !=
+                      datetime.datetime.strptime(date, UTC_DATE_STRING).date()):
```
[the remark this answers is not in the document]
ialokim: Done.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+    global base_data
+
+    canteen = LazyBuilder()
+    with urlopen(url) as response:
+        data = json.loads(response.read().decode())
+
+    for day in data['days']:
+        date = day.get('date')
+
+        if today and (datetime.datetime.now().date() !=
```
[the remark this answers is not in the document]
ialokim: Thanks, changed it accordingly.

parsers/saarland.py
```diff
@@ -0,0 +1,199 @@
...
+
+def parse_url(url, today=False):
+    global base_data
+
+    canteen = LazyBuilder()
+    with urlopen(url) as response:
+        data = json.loads(response.read().decode())
+
+    for day in data['days']:
+        date = day.get('date')
```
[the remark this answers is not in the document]
ialokim: You're right, thanks!

parsers/saarland.py
ialokim: Okay, I've changed it to a little sub-function. However, I would like to keep the separating by an "und" for the last item.

parsers/saarland.py
ialokim: Okay, I've changed it to throw an error instead.

parsers/saarland.py
ialokim: Humm, okay, so you suggest hardcoding the different locations once so that they cannot be updated automatically when the `BaseData` has been updated? Or am I missing some way this would be possible without loading the `BaseData` in the main scope?

ialokim:
I've now addressed most of the issues, thanks a lot for your review!

Please see my comment concerning the loading of the base data inside the main scope.

parsers/saarland.py
klemens: > […] hardcoding the different locations once […]

Yes, I think this is the best solution for now.

> Or am I missing some way this would be possible without loading the BaseData in the main scope?

I don't think this is possible right now, but we can certainly discuss adding a feature that allows fetching the canteens dynamically, maybe together with support for actually creating the new canteens at openmensa.org automatically (through extended metadata).

ialokim: Okay, I've changed it now.

parsers/saarland.py
```diff
@@ -0,0 +1,223 @@
...
+        if loc not in LOCATIONS:
+            # Found an unknown location
+            # Please consider updating the parser!
+            raise RuntimeError(
+                'Unknown location: %s (displayName: %s)' %
+                (loc, data['locations'][loc]))
+    for role in data['priceTiers']:
+        if role not in ROLES:
+            # Found an unknown price tier
+            # All prices for this price tier will be ignored.
```
klemens: This line of the comment is no longer accurate.

parsers/saarland.py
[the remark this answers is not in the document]
klemens: I think we can leave it for now, if this is typically the location and change it if there are actually other types of descriptions someday.

parsers/saarland.py
```diff
@@ -0,0 +1,223 @@
...
+            counter_name = counter['displayName']
+            counter_description = counter['description']
+            counter_hours = counter.get('openingHours')
+
+            for meal in counter['meals']:
+                if 'knownMealId' in meal:
+                    # This is meant to allow recognizing recurring meals,
+                    # for features like marking meals as favorites.
+                    # Up to now, not really used in the mensaar.de API.
+                    print('knownMealId: %s' % meal['knownMealId'])
```
klemens: I's fine if you want to keep it, but please print to stderr with `file=sys.stderr`.

parsers/saarland.py
```diff
@@ -0,0 +1,223 @@
...
+            # All prices for this price tier will be ignored.
+            # Please consider updating the parser!
+            raise RuntimeError(
+                'Unknown price tier: %s (displayName: %s)' %
+                (role, data['priceTiers'][role]))
+        else:
+            base_data['roles'][role] = ROLES[role]
+
+
+base_data = {}
```
klemens: Is there is specific reason for storing this data in a global variable? If not I would prefer just returning the data from the function and passing it to the other functions as needed.

parsers/saarland.py
```diff
@@ -0,0 +1,223 @@
...
+                    build_notes(meal['notices'], meal['components']))
+
+                meal_prices = {}
+                if 'prices' in meal:
+                    prices = meal['prices']
+                    for role in prices:
+                        if role in base_data['roles']:
+                            meal_prices[base_data['roles'][role]] = prices[role]
+                        else:
+                            print('Ignoring price for unknown role %s.' % role)
```
klemens: I guess this cannot happen anymore, as we now abort the parser when there are new price tiers, so we could remove the check and print. If you want to keep the print, please also print to stderr like above.

parsers/saarland.py
ialokim: Removed it.

parsers/saarland.py
ialokim: Okay, I've changed it.

parsers/saarland.py
ialokim: No, there is no reason. I've changed it.

parsers/saarland.py
ialokim: Removed the check.

parsers/saarland.py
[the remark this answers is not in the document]
ialokim: I've added a little comment explaining this.

klemens:
Thanks again! I just merged the parser. It should become available under http://omfeeds.devtation.de/saarland/sb/today.xml etc. really soon.

Do you have an openmensa.org account (or want to create one) and add the canteens? If so, you have to add a new "parser" there and add the canteens using the same name as in the parser here. You can then add the index url (http://omfeeds.devtation.de/saarland/index.json) to the "parser" and press update, which will add the right feed urls automatically.

ialokim:
> Thanks again! I just merged the parser. It should become available under http://omfeeds.devtation.de/saarland/sb/today.xml etc. really soon.

Thanks a lot!

> Do you have an openmensa.org account (or want to create one) and add the canteens? If so, you have to add a new "parser" there and add the canteens using the same name as in the parser here. You can then add the index url (http://omfeeds.devtation.de/saarland/index.json) to the "parser" and press update, which will add the right feed urls automatically.

I did this, but I think there might be some problem, because it does not show feeds in the parser details. Also, if I click the "update sources with index url" button, it says "5 new sources, 0 sources added". Opening the index url right away in the browser gives an error page. Not sure what I'm doing wrong?

klemens:
> Opening the index url right away in the browser gives an error page

This is due to the server returning the wrong mime-type. You can view the page source to work around that problem.

> I did this, but I think there might be some problem, because it does not show feeds in the parser details

Have you manually added the canteens (including address etc)? Currently they cannot be automatically added by the index url, because the metadata returned only contains information about the feeds and not the canteens itself.

ialokim:
> Have you manually added the canteens (including address etc)? Currently they cannot be automatically added by the index url, because the metadata returned only contains information about the feeds and not the canteens itself.

Okay, got it. But it seems the omfeeds.devtation.de server is offline now.

Btw, it seems they added two more canteens, I should open a new PR adding these, right?

klemens:
> But it seems the omfeeds.devtation.de server is offline now.

It is online again.

> Btw, it seems they added two more canteens, I should open a new PR adding these, right?

Yes, please do so!

ialokim:
> It is online again.

Okay, but there still seems to be an error, loading http://omfeeds.devtation.de/saarland/sb/today.xml gives a 502 Bad Gateway page. :thinking:

klemens:
It fails because of the new locations:
> RuntimeError: Unknown location: htwcas (displayName: {'displayName': 'Mensa HTW Saar CAS', 'description': '…'})

Maybe we shouldn't fail hard when there are new locations.

ialokim:
> It fails because of the new locations

Good point. See #83.

klemens:
Merged! Should be up *soon*.

PR #78, opened 2018-04-19
Add simple metadata endpoint to HandleSource parsers

jgraichen:
```
This commit adds a metadata.xml handler for HandleSource. It will generate
a plain metadata XML with a default full and today feed URLs.

Metadata will only include the feed definitions and no canteen information to
avoid overriding data. This means (new) canteens *cannot* be created by adding
metadata targets only.

Closes openmensa/openmensa#35
```

This should fix openmensa/openmensa#35 while keeping the Index-URLs for parses build with HandleSource. The next update from OpenMensa should fetch the new metadata and update existing canteens the default today and full feeds.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -5,7 +5,7 @@ import json
 
 from bs4 import BeautifulSoup
 
-from pyopenmensa.feed import LazyBuilder
+from pyopenmensa.feed import LazyBuilder, Feed
 
 
 class Request(object):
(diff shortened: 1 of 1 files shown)


utils.py
```diff
@@ -168,7 +168,39 @@ def __init__(self, name, parser, handler, args=[], kwargs={}, default_feed='full
         self.args = args
         self.kwargs = kwargs
 
+    def metadata(self, request):
+        meta = LazyBuilder(version='0.1.0')
```
klemens: Should we use the actual version of the parser, which leaves of the version tag in the xml in most cases: `LazyBuilder(version=self.parser.version)`? Or does this only work when a version is present?

utils.py
```diff
@@ -168,7 +168,39 @@ def __init__(self, name, parser, handler, args=[], kwargs={}, default_feed='full
...
+
+        meta.feeds.append(Feed(
+            name='today',
+            hour='8-14',
```
klemens: nit: I know that this has been the default for a while, but it seem weird to fetch the full *and* today feed at 8 o'clock. I think `9-14` would make more sense.

jgraichen: Afaik not all parsers include "today" in the "full" feed. I'd therefore like to leave the defaults for now.

It probably should be made configurable anyway.

klemens: Ok, then let's leave this for now.

PR #79, opened 2018-04-26
Added Parser for most of the Mensas from Studentenwerk Thueringen

Anarchoschnitzel:
Added Parser for
- Mensa Ehrenberg
- Cafeteria Ehrenberg
- Nanoteria
- Roentgenmensa
from TU Ilmenau.
http://www.stw-thueringen.de/deutsch/mensen/einrichtungen/ilmenau/mensa-ehrenberg.html
http://www.stw-thueringen.de/deutsch/mensen/einrichtungen/ilmenau/cafeteria-mensa-ehrenberg.html
http://www.stw-thueringen.de/deutsch/mensen/einrichtungen/ilmenau/cafeteria-nanoteria.html
http://www.stw-thueringen.de/deutsch/mensen/einrichtungen/ilmenau/cafeteria-roentgenbau.html
My current implementation marks days in the past as closed. Please let me know if that is not the desired behavior and I will change it :)


Y0hy0h:
Hey, thanks for your PR!

> My current implementation marks days in the past as closed.

I'm not fully sure what you mean, and I understand your code enough to verify if you did it correctly. OpenMensa keeps a history of all parsed days. I assume you mean when today's menu (say Wednesday, 2018-05-02) contains entries for the past week or the day before (2018-04-23, or 2018-05-01), then the correct behavior is to just output them as well. So the feed will simply contain all the days from the menu, and OpenMensa will update the "old" days accordingly.

I will just quickly go through your code and point out a few general things that you could improve. ;)

Anarchoschnitzel:
Hi Y0hy0h, what I mean by past days as closed is the following:
The websites I pull the information from only shows the menus from the current day on and not from days in the past. These days are currently marked as closed.
Example:
The parser is called on a Wednesday and returns alle meals from Wednesday, Thursday and Friday.
It then determines Saturday and Sunday as closed.
As no meals were parsed for Monday and Tuesday (which are in the past from the current day) these days are marked as closed.
Do you think it would be better to return no info about these day in the past at all?

parsers/thueringen.py
```diff
@@ -0,0 +1,103 @@
...
+					canteen.addMeal(current_date.date(), 'Hauptgerichte', main_dish, notes, costs,
+									None)
+				if today:
+					return canteen.toXMLFeed()
+			for i in range(2, 9):
+				if i not in days_open:
+					closed_date = mensa_start_date + timedelta(days=i-2)
+					canteen.setDayClosed(closed_date.date())
+		except Exception:
+			continue
```
Y0hy0h: This is dangerous, because you're swallowing _all_ exceptions. Prefer making the `try` block as short as possible and only `except`ing _exactly_ the errors you are interested in. Or even better, replace it with a clear check (using e. g. if). That way it becomes clear what you are checking, and you do not accidentally swallow other exceptions.

parsers/thueringen.py
```diff
@@ -0,0 +1,103 @@
+from urllib.request import urlopen
+from bs4 import BeautifulSoup as parse
+import re
+from utils import Parser
+from datetime import datetime, timedelta
+from pyopenmensa.feed import LazyBuilder
+
+calendar_week_regex = re.compile('\d{4}-\d{1,2}$')
```
Y0hy0h: Prefer declaring variables close to where they are used. Readability trumps performance, since you can always optimize readable code, when you've determined that the performance is important.

parsers/thueringen.py
```diff
@@ -0,0 +1,103 @@
...
+	canteen = LazyBuilder()
+	content = urlopen(url).read()
+	document = parse(content, 'lxml')
+	employees_fee = 1.8
+	guests_fee = 3.2
+	fees = document.find_all('p', {'class': 'MsoNormal'})
+	for fee_candidate in fees:
+		fee_string = fees_regex.findall(fee_candidate.text)
+		for s in fee_string:
+			amount_strings = amount_regex.findall(s)
```
Y0hy0h: This is deeply nested. Prefer extracting a function and name it after what happens here.

parsers/thueringen.py
```diff
@@ -0,0 +1,103 @@
...
+			mensa_start_date = datetime.strptime(start_date_str, '%d.%m.%Y')
+			days = document.find_all('div', id=days_regex)
+			days_open = []
+			for day in days:
+				day_id = int(day['id'][-1:])
+				days_open.append(day_id)
+				current_date = mensa_start_date + timedelta(days=day_id-2)
+				meals_data = day.find_all('tr')
+				for meal_data in meals_data:
+					meals = meal_data.find_all('td')
```
Y0hy0h: Same here: Extract a function. Small functions are awesome, because you can quickly understand them, whereas such a big function is difficult to read.

Anywhere you _can_ reasonably extract a function, do it. :)

parsers/thueringen.py
Anarchoschnitzel: I moved the variables to the functions they are used at except one which is used at multiple places

parsers/thueringen.py
Anarchoschnitzel: I extracted and split the parsing into 3 different functions

parsers/thueringen.py
Anarchoschnitzel: I extracted and split the parsing into 3 different functions

parsers/thueringen.py
Anarchoschnitzel: I reworked that part to get rid of the long and nested `try` block. 
Thanks for your recommendations 👍

Anarchoschnitzel:
I took care of the changes you requested.
Thanks for your feedback 👍

Y0hy0h:
> Do you think it would be better to return no info about these day in the past at all?

Yes, that's what I would do. ;)

Anarchoschnitzel:
> Yes, that's what I would do. ;)
This is already changed in the commits that also addressed the changes you requested.

parsers/thueringen.py
```diff
@@ -0,0 +1,134 @@
...
+				costs['employee'] = costs['student'] + employees_fee
+			if guests_fee is not -1:
+				costs['other'] = costs['student'] + guests_fee
+			if today:
+				if current_date.day is datetime.today().day:
+					canteen.addMeal(current_date.date(), 'Hauptgerichte', main_dish, notes, costs,
+									None)
+			else:
+				canteen.addMeal(current_date.date(), 'Hauptgerichte', main_dish, notes, costs,
+								None)
```
klemens: I think you should check if `today and current_date != datetime.today()` before parsing the day and just `continue` if it is the case.

parsers/thueringen.py
```diff
@@ -0,0 +1,134 @@
...
+
+amount_regex = re.compile('\d{1,2},\d{1,2}')
+
+
+def parse_start_date(document):
+	week_start_end_date_regex = re.compile('\d{1,2}\.\d{1,2}\.\d{4}')
+	calendar_week_regex = re.compile('\d{4}-\d{1,2}$')
+	options = document.find_all('option', value=calendar_week_regex)
+	assert len(options) > 0
+	options_filtered = list(filter(lambda x: x.has_attr('selected'), options))
```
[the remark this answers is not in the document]
Anarchoschnitzel: Good to know, I am fairly new to BS and had my problems with that one ;-)

parsers/thueringen.py
```diff
@@ -0,0 +1,134 @@
...
+	content = urlopen(url).read()
+	document = parse(content, 'lxml')
+	fees = parse_fees(document)
+	if len(fees) is 2:
+		employees_fee = fees[0]
+		guests_fee = fees[1]
+	try:
+		mensa_start_date = parse_start_date(document)
+	except AssertionError:
+		return canteen.toXMLFeed()
```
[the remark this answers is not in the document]
Anarchoschnitzel: With your feedback above, I changed the parse_start_date function so it should not throw an exception (unless something goes **really** wrong and as you stated in that case we want the parser to fail).
As a result the parse_date function is also a lot shorter than before :)

parsers/thueringen.py
```diff
@@ -0,0 +1,134 @@
...
+			if guests_fee is not -1:
+				costs['other'] = costs['student'] + guests_fee
+			if today:
+				if current_date.day is datetime.today().day:
+					canteen.addMeal(current_date.date(), 'Hauptgerichte', main_dish, notes, costs,
+									None)
+			else:
+				canteen.addMeal(current_date.date(), 'Hauptgerichte', main_dish, notes, costs,
+								None)
+	for i in range(min(days_open), 9):
```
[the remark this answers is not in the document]
Anarchoschnitzel: I modified this line into `for day_id in range(min(days_open or [9]), 9):` so if days_open is empty the loop will be skipped

Anarchoschnitzel:
Hi there, I know you are pretty busy atm, but I just wanted to check if there is any progress regarding the PullRequest :)

parsers/thueringen.py
```diff
@@ -0,0 +1,148 @@
...
+
+def parse_meals(day):
+	ingredients_regex = re.compile('Inhalt:.*')
+
+	meals = []
+	meals_t_rows = day.find_all('tr')
+
+	for meal_data in meals_t_rows:
+		meal_t_datas = meal_data.find_all('td')
+		if len(meal_t_datas) is not 3:
```
klemens: Please use `!=` instead of `is not`.

parsers/thueringen.py
```diff
@@ -0,0 +1,148 @@
...
+			continue
+
+		notes = [meal_t_datas[0].text]
+
+		ingredients = ingredients_regex.findall(meal_t_datas[1].text)
+		if len(ingredients) > 0:
+			notes.append(ingredients[0].strip())
```
klemens: You should add a single note for every ingredient. Also please parse the table at the bottom of the page to write out the abbreviations in full. There are some helpers for this in https://github.com/mswart/pyopenmensa/blob/7b6e7a70ab63530af27b985019f28863f8b8f368/feed.py#L182-L246 that can parse ingredient tables and automatically replace the abbreviations. I can also help you out with this if you want.

parsers/thueringen.py
```diff
@@ -0,0 +1,148 @@
...
+		main_dish = ingredients_regex.sub('', meal_t_datas[1].text).strip()
+
+		students_fee_string = amount_regex.findall(meal_t_datas[2].text)
+		if len(students_fee_string) != 1:
+			continue
+
+		students_fee = float(students_fee_string[0].replace(',', '.'))
+		costs = {'student': students_fee}
+
+		meals.append((main_dish, notes, costs))
```
klemens: You can just `yield` the tuples instead of storing them in a list and then returning the list.

parsers/thueringen.py
```diff
@@ -0,0 +1,148 @@
...
+	return meals
+
+
+def parse_url(url, today=False):
+	days_regex = re.compile('day_\d$')
+
+	canteen = LazyBuilder()
+
+	content = urlopen(url).read()
+	document = parse(content, 'lxml')
```
klemens: You are only parsing the current week currently, which means that the full fead does not contain the meals for Monday on Sunday, but last Friday instead, which is not very useful.

To get the next week, you have to send a request with the `selWeek=2018-25` parameter, where 25 is the calender week. It seems you can also get a list view using `selView=liste` which contains the explicit date, which might be easier to parse that the tabbed view.

The website uses POST-requests, but it seems GET also works: https://www.stw-thueringen.de/deutsch/mensen/einrichtungen/ilmenau/mensa-ehrenberg.html?selWeek=2018-25&selView=liste

Anarchoschnitzel:
First of all, thanks for your feedback! I haven't worked much with python so far so every bit of feedback that helps me to enhance my knowledge of the language is great 👍 
I have only one small question left regarding the changes I implemented: as I am now fetching the meals for the canteens for the current week + 2 - as you suggested - it sometimes happens that the meals for 2 weeks in advance are not fully present on the Studentenwerks pages. This results in some week tabs not being there which my parser interprets as "canteen closed on that day" (which is in theory possible during the week etc. for holidays). Now my question is: if the OpenMensa server calls my parser later again and the data for the week is there so days that were previously returned as closed are returned as open and with meals, will it overwrite the existing information for "closed on that day" or ignore it?

Thanks for your help in advance :)

klemens:
> if the OpenMensa server calls my parser later again and the data for the week is there so days that were previously returned as closed are returned as open and with meals, will it overwrite the existing information for "closed on that day" or ignore it?

I am not 100% sure, but I think that newer data always overwrites older data. Alternatively, you could just return no data for those days and not bother with the "canteen closed" status at all.

> I referenced the wrong issue in my commit message by accident, sorry for that ;)

Don't worry, we can clean up the commit history before merging!

Anarchoschnitzel:
Allrighty, days that don't appear on the websites tabs are now being omitted 👍

parsers/thueringen.py
```diff
@@ -0,0 +1,199 @@
...
+	ingredients = []
+	for ingredient in ingredients_list:
+
+		if digits.get(ingredient) is not None:
+			ingredients.append(digits[ingredient][1:])
+
+		if chars.get(ingredient) is not None:
+			ingredients.append(chars[ingredient][1:])
+
+	return ingredients
```
klemens: You can do this whole function in one line using list comprehensions if you want to:
```python
notes = [legend[note] for note in notes if note in legend]
```

parsers/thueringen.py
```diff
@@ -0,0 +1,199 @@
...
+	for meal_data in meals_t_rows:
+		meal_t_datas = meal_data.find_all('td')
+		if len(meal_t_datas) != 3:
+			continue
+
+		category = [meal_t_datas[0].text]
+		if len(category) == 1:
+			category = category[0]
+		else:
+			category = 'Hauptgerichte'
```
klemens: `len(category) == 1` will always be true, because you just created a list with length 1 above? Did you mean to check if the string was empty?

parsers/thueringen.py
```diff
@@ -0,0 +1,199 @@
...
+
+		notes = []
+		ingredients = ingredients_regex.findall(meal_t_datas[1].text)
+
+		if len(ingredients) > 0:
+			ingredients_list = ingredients[0].strip('Inhalt: ').strip('\r').split(',')
```
klemens: Please use `strip()` instead of `strip('\r')`. Also you should strip the resulting list items (maybe combined with the list comprehension from above?), because there may be some spaces in there sometimes.

parsers/thueringen.py
```diff
@@ -0,0 +1,199 @@
...
+
+def parse_url(url, today=False):
+	year = datetime.now().isocalendar()[0]
+	week = datetime.now().isocalendar()[1]
+
+	days_regex = re.compile('day_\d$')
+
+	canteen = LazyBuilder()
+
+	for i in range(0, 3):
```
klemens: You should probably only fetch the current week when the today parser is requested.

parsers/thueringen.py
```diff
@@ -0,0 +1,199 @@
...
+			digits, chars = parse_ingredients(document)
+			meals = parse_meals(day, digits, chars)
+			for meal in meals:
+				main_dish, notes, costs, category = meal
+
+				if employees_fee is not None:
+					costs['employee'] = costs['student'] + employees_fee
+				if guests_fee is not None:
+					costs['other'] = costs['student'] + guests_fee
+				if today and current_date != datetime.today():
```
klemens: `./parse.py thueringen il-ehrenberg today.xml` currently returns an empty result for me, because this comparison between dates is not actually between dates, but between datetimes. It should be `current_date != date.today()` and `parse_start_date` should return a date instead of a datetime (just use `.date()` before returning).

Anarchoschnitzel:
I implemented the proposed changes.
I also found a ( I think better / more reliable) way to fetch meals for upcoming weeks, as I now parse the <selection> element from the document which has all available weeks for the future as options. This should also remove the necessary of bothering with leading zeroes in the ?selWeek parameter

parsers/thueringen.py
```diff
@@ -0,0 +1,169 @@
...
+
+def parse_ingredients(document):
+	groups_regex = re.compile('([\w\d*]+):\s+(.*)')
+	groups = dict()
+
+	for table in document.select('div.kontextbox > table:nth-of-type(1)'):
+		for s in table.stripped_strings:
+			if groups_regex.search(s):
+				split = s.split(':')
+				groups[split[0]] = split[1]
```
klemens: The object returned by `search` already contains the matched groups (the parenthesis in the regex), which can be accesses using `group(1)` and `group(2)`, so you don't need `split`. You should also strip the description here instead of in the list comprehension in line 84.

parsers/thueringen.py
```diff
@@ -0,0 +1,169 @@
...
+		meal_t_datas = meal_data.find_all('td')
+		if len(meal_t_datas) != 3:
+			continue
+
+		category = meal_t_datas[0].text
+		notes = []
+		ingredients = ingredients_regex.findall(meal_t_datas[1].text)
+
+		if len(ingredients) > 0:
+			ingredients_list = ingredients[0].strip('Inhalt: ').strip().split(',')
```
klemens: Sorry, I missed this in the last review. `strip('Inhalt: ')` probably does not do what you think it does. (I suggest just using `[8:]` or a regex)

parsers/thueringen.py
```diff
@@ -0,0 +1,169 @@
...
+def parse_url(url, today=False):
+
+	days_regex = re.compile('day_\d$')
+
+	canteen = LazyBuilder()
+
+	content = urlopen(url).read()
+	document = parse(content, 'lxml')
+
+	available_weeks = parse_available_weeks(document)
```
klemens: This fetches the current week's site twice. I suggest skipping the current week and using the already available `document` for adding the meals. It probably makes sense to split the `parse_url` function into two functions, one which fetches the sites and checks which are available and the other does the actual parsing (what's currently in the for loop).

parsers/thueringen.py
```diff
@@ -0,0 +1,169 @@
...
+
+			meals = parse_meals(day, groups)
+			for meal in meals:
+				main_dish, notes, costs, category = meal
+
+				if employees_fee is not None:
+					costs['employee'] = costs['student'] + employees_fee
+				if guests_fee is not None:
+					costs['other'] = costs['student'] + guests_fee
+				if today and current_date != date.today():
```
klemens: This still fetches all sites, even when just the first one is needed. You need another check that only fetches the current site for today feeds (see also the comment above).

parsers/thueringen.py
```diff
@@ -0,0 +1,169 @@
...
+
+	option = document.find('option', value=calendar_week_regex, selected=True)
+
+	start_date_str = week_start_end_date_regex.search(option.text)
+
+	return datetime.strptime(start_date_str.group(), '%d.%m.%Y').date()
+
+
+def parse_available_weeks(document):
+
```
klemens: NIT: I find empty lines after function definitions of for loops (see below) a little bit confusing. They also seem to confuse the git diff algorithm. :wink:

Anarchoschnitzel:
I implemented the proposed changes - thanks for your feedback 👍

Anarchoschnitzel:
I just found out my solution might not work properly yet. If accessed on weekends / Sundays the Studentenwerk page does not auto select a week which currently crashes the parser on those days. I am working on a fix.

Anarchoschnitzel:
Should be fixed now

klemens:
Sorry for – once again – taking so long. The changes look good, I just pushed a small change to replace the `assert` with a normal check for `Null`, as exceptions should (normally) not be used for control flow.

I will now merge this (sqashed into one commit) into the master branch, which means that it will become available at `http://omfeeds.devtation.de/thueringen/*location*/{full,today}.xml` within a day or so. Then you can request an update or the transfer of the parser to your account, if you want. Otherwise, I can also do this.

Thanks again for writing the parser and following my (sometimes nitpicky) review! :grin:

Anarchoschnitzel:
Thanks for the update! 
If I find the time I will head over to the openmensa project page and request an update in the upcoming days.

Again thanks for the continuous feedback, was really fun writing this parser and learning a few new things and best practices about python 👍

klemens:
Soooo, I just moved all of the already existing canteens into one parser (they were separated by city) and adapted their names to your new naming scheme. After adding the [new index-url](https://omfeeds.crpt.de/thueringen/index.json) of my server, they started to work immediately. :wink: 

* https://openmensa.org/c/147
* https://openmensa.org/c/148
* https://openmensa.org/c/151
* https://openmensa.org/c/150
* https://openmensa.org/c/149
* https://openmensa.org/c/74
* https://openmensa.org/c/75
* https://openmensa.org/c/76

I did not add the other canteens yet, mainly because I wanted to automate it (see the halle parser and my changes to it today). My plan is as follows: The city overview page (eg [Jena](https://www.stw-thueringen.de/deutsch/mensen/einrichtungen/jena/index.html)) contains a list of canteens including their addresses. It also contains a hidden form, which contains geo-coordinates for all canteens (used for the map display). This way we should be able to generate the complete metadata info and avoid copying it by hand into openmensa (which of course would be much faster, but missing all the fun :grin:).

klemens:
I implemented the metadata extraction in 90b95c9 and fixed a small issue, where an [empty meal name](https://www.stw-thueringen.de/deutsch/mensen/einrichtungen/jena/cafeteria-eah.html) would lead to an exception, because they are not allowed. Now, empty meals are just skipped: fddcca4

The extraction worked great for all but three canteens:

* https://openmensa.org/c/833
* https://openmensa.org/c/834
* https://openmensa.org/c/835
* https://openmensa.org/c/836
* https://openmensa.org/c/837
* https://openmensa.org/c/838
* https://openmensa.org/c/839
* https://openmensa.org/c/840
* https://openmensa.org/c/841
* https://openmensa.org/c/842
* https://openmensa.org/c/843
* https://openmensa.org/c/844
* https://openmensa.org/c/845

These three were missing in the map display, so I created them manually:

* https://openmensa.org/c/846
* https://openmensa.org/c/847
* https://openmensa.org/c/848

I did some quick sanity checks (the coordinates are in the right city), but it would be great if you could also take a look!

Issue #89, opened 2018-12-12
Upload of new packages from travis to apt.dxtt.de fails with 406 Not Acceptable


klemens:
@mswart
The travis runs for the last three commits on master ([cedd712](https://travis-ci.org/mswart/openmensa-parsers/builds/466700882), [2eea58b](https://travis-ci.org/mswart/openmensa-parsers/builds/467173421), and [3fd2f37](https://travis-ci.org/mswart/openmensa-parsers/builds/467181917)) failed to upload the created package to apt.dxtt.de with the following error:
```
Resolving apt.dxtt.de (apt.dxtt.de)... 46.38.231.137, 2a03:4000:20:39::34
Connecting to apt.dxtt.de (apt.dxtt.de)|46.38.231.137|:80... connected.
HTTP request sent, awaiting response... 406 Not Acceptable
2018-12-12 20:07:41 ERROR 406: Not Acceptable.
```

klemens:
I now have access to all parsers on openmensa.org, so I am now hosting the parsers at my own server: https://omfeeds.crpt.de

I have already switched over hannover, dresden, and marburg and @ialokim changed saarland. If everything works smoothly tomorrow, I will switch over the rest of the parsers.

This will also finally allow us to fix the [thueringen](https://github.com/mswart/openmensa-parsers/pull/79) (@Anarchoschnitzel) parser and add the new canteens to [wuerzburg](https://github.com/mswart/openmensa-parsers/pull/82) (@pstiegele) and [muenchen](https://github.com/mswart/openmensa-parsers/pull/85) (@FloEdelmann, @chneukirchen). Sorry again to all for the extremly long delays :disappointed:.

klemens:
Everything seems to work fine with the new server! I switched over some more canteens: wuerzburg, muenchen, duesseldorf, halle (metadata currently broken)

I added the new canteens for wuerzburg and munechen and started preparing the ones for thueringen (not done yet, need to collect all addresses and geo locations).

I also added a [list](https://github.com/mswart/openmensa-parsers/wiki) to the wiki that contains links to all parsers on openmensa.org. Note that this is mainly useful for myself currently, as you cannot access these without being the owner or an admin.

jgraichen:
@klemens You can access a parser as an administrator through the canteen page. There is a link "Parser öffnen" in the sidebar. It's not very elegant but best we have right now.

[...]

PR #90, opened 2019-01-07
Fix dresden parser

hesstobi:
Studentenwerk Dresden launched a redesign of their webpage, thus the parser was not working anymore.

### DIFF
--- a/build_scripts/test-urls.txt
+++ b/build_scripts/test-urls.txt
@@ -15,3 +15,4 @@ http://localhost:9090/wuerzburg/austrasse/full.xml
 http://localhost:9090/erlangen_nuernberg/ansbach/full.xml
 http://localhost:9090/rostock/mensa-sued/full.xml
 http://localhost:9090/saarland/sb/full.xml
+http://localhost:9090/dortmund/tu-hauptmensa/full.xml


hesstobi:
I don't like failing tests...therefore fixed marburg

parsers/dresden.py
```diff
@@ -6,53 +6,90 @@
 
 from pyopenmensa.feed import LazyBuilder, extractDate
 
-price_regex = re.compile('(?P<price>\d+[,.]\d{2}) ?€')
+price_regex = re.compile(r'(?P<price>\d+[,.]\d{2}) ?€')
+heading_regex = re.compile(r'^h[1-6]$')
```
klemens: Do you think they will randomly change the header type? :grin:

parsers/dresden.py
```diff
@@ -6,53 +6,90 @@
...
+                canteen.setDayClosed(date)
+                continue
+
+            # Iterate over the list-group-item within the card which are used
+            # for individual meals
+            for meal in day_card.find_all(class_='list-group-item'):
+
+                try:
+                    name = meal.find(name='span').span.text
+                except AttributeError:
```
klemens: You could just check if find returns None, also the second span seems unnecessary.

parsers/dresden.py
```diff
@@ -6,53 +6,90 @@
...
+                    # No Name...no meal
+                    continue
+
+                if ': ' in name:
+                    category, name = name.split(': ', 1)
+                else:
+                    category = 'Angebote'
+
+                try:
+                    notes = [img['alt'] for img in name.find_next(name='span').find_all(name='img')]
```
klemens: This does not work, as name is a string at this point.

parsers/dresden.py
```diff
@@ -6,53 +6,90 @@
...
+                    # 2nd span -> no notes
+                    notes = []
+
+                if '* ' in name:
+                    name, note = name.split('* ', 1)
+                    notes.append(note)
+
+                try:
+                    prices = price_regex.findall(meal.strong.text)
+                except AttributeError:
```
klemens: Please use `if meal.strong is not None` or equivalent.

parsers/marburg.py
```diff
@@ -1,61 +1,73 @@
...
+def parse_dish(dish, canteen):
+    date = extractDate(dish['data-date'])
+
+    name = dish.find(class_='neo-menu-single-title')
+    try:
+        notes = set(x['title'] for x in name.find_all(name='abbr'))
+    except AttributeError:
+        # No Name
+        return
+    name = re.sub(notes_regex, '', name.get_text(strip=True))
```
klemens: You could also use `name.text.strip()`, which is shorter. :wink:

Also you should check if the name is empty afterwards, because empty meal names are not allowed and result in: `ValueError: Meal names must not be empty`

parsers/marburg.py
```diff
@@ -1,61 +1,73 @@
...
+
+    try:
+        category = dish.find(class_='neo-menu-single-type').text
+    except AttributeError:
+        # No categprie this is maby a side
+        category = 'Beilagen: ' + dish.find_previous(name='h2').text.capitalize()
```
klemens: This seems brittle and should not end parsing if it fails.

parsers/marburg.py
```diff
@@ -1,61 +1,73 @@
...
+
+    try:
+        price = dish.find(class_='neo-menu-single-price').text
+        prices = price_regex.findall(price)
+    except AttributeError:
+        # No Price
+        prices = []
```
klemens: The default should be `{}`.

parsers/marburg.py
```diff
@@ -1,61 +1,73 @@
...
-        for tr_menu in day_table.tbody.find_all('tr'):
-            category = tr_menu.find_all('td')[0].text.strip()
-            name = tr_menu.find_all('td')[1].text.replace('\r\n', ' ').strip()
-            canteen.addMeal(date, category, name, [], tr_menu.find_all('td')[2].text)
-
-
-def parse_url(url, mensa, *weeks, today):
+from pyopenmensa.feed import LazyBuilder, extractDate
+
+heading_regex = re.compile(r'^h[1-6]$')
```
klemens: This is not used.

parsers/marburg.py
```diff
@@ -1,61 +1,73 @@
...
+notes_regex = re.compile(r'\((\d+,?\s*)*\)\s*')
+price_regex = re.compile(r'(?P<price>\d+[,.]\d{2}) ?€')
+
+roles = ('student',)
```
klemens: You can use a simple array here: `['student']`

parsers/marburg.py
```diff
@@ -1,61 +1,73 @@
...
+    except AttributeError:
+        # No disches
+        pass
+    try:
+        sides = document.find(class_='neo-menu-single-modals').find_all(name='tr', attrs={"data-canteen": data_canteen})
+    except AttributeError:
+        # No sides
+        pass
+
+    [parse_dish(dish, canteen)for dish in dishes + sides]
```
klemens: As you are not returning anything from the function, a simple for loop is more idiomatic:
```suggestion
    for dish in dishes + sides:
        parse_dish(dish, canteen)
```

parsers/marburg.py
```diff
@@ -1,61 +1,73 @@
...
+
+
+def parse_url(url, data_canteen, today=False):
```
klemens: NIT: As the function is now basically only calling `parse_week`, those two could be merged.

parsers/marburg.py
hesstobi: What is wrong with the tuple?

klemens: Technically, nothing. It just feels a little weird to use a tuple to pass in a ordered list of roles. But I don't care strongly.

klemens:
Thanks! I squashed the commits into one for each parser and merged them. There seems to be a problem with hannover (I will take a look at it) and #89 is still a problem, so it might take a while before the changes are visible unfortunately.

klemens:
So, hannover is fixed and I now host the parsers myself. So dresden and marburg *should* start working again tomorrow.

PR #91, opened 2019-01-23
Saarland: bump to new api version and support two more locations

ialokim:


klemens:
We currently have some problems updating the parser: #89 However, as you are the owner of the parser at opemensa.org, I can offer you to host the parser at my own server for now. You could then change to index url so that it points to my server.

ialokim:
> host the parser at my own server for now
That would be great! Just provide me the new url and I would change it on openmensa.

klemens:
Here you go: https://omfeeds.crpt.de/saarland/index.json

I guess it is enough to add the new canteens and update the index url, then press `Aktualisiere Quellen mittels Index-URL`.

I you notice anything unusual with the new server, just ping me.

ialokim:
Thanks a lot, everything works as expected (at least for now). Was a lot easier to switch to your server than I've thought before.

PR #88, opened 2018-11-09
New Parser for TU-Dortmund and FH-Suedwestfalen

macintosh-HD:
I wrote a parser for the canteens that the Studierendenwerk Dortmund is working for. The parser can parse all canteens of the TU-Dortmund, FH Dortmund, ISM Dortmund, FH Suedwestfalen and FernUni Hagen.

PR #94, opened 2019-02-13
Dortmund: fix error when parsing meals with no name

macintosh-HD:
Like @mswart mentioned in #88 I totally disregarded the fact that meal names are required.
Now a check is performed right before the meal is added.


klemens:
> Like [@]mswart mentioned

Almost :wink:.

I manually cherry-picked your commit into master (87f2540e19fff3788201547a1cd11a9b38dc98cb), as there seems to be some problem with your PR [1]. I also removed the (accidental?) submodule update, as this should be it's own commit. Thanks again!

[1]: As I combined all commits into a single one when merging your original PR, your master branch and the upstream master branch are no longer compatible, which is why all these commits are also included in this PR. To avoid this, you should always create a new branch for your changes (e.g. one per PR) and use your master branch only to get the newest commits from upstream and base your branches on.

macintosh-HD:
> Almost 😉

🤦‍♂️  Sorry for that! 😃 

Thanks for the advise I will remember that for future contributions.

PR #97, opened 2019-05-13
Fix dresden parser

hesstobi:
New css class `swdd-link-list-item` for meals.


klemens:
Thanks! I updated all canteens on openmensa.org, meals are found again.

PR #102, opened 2019-10-26
Aachen: Add note for vegetarian and vegan meals

manorom:
This adds additional notes if the meal parsed is vegetarian (OLV) or
vegan. On the website, vegetarian and vegan meals are not marked with
notes like allergens but with a css class.


klemens:
Thanks! Merged and updated on openmensa.org.

PR #104, opened 2019-11-03
Vegan/Vegetarian notes for munich mensa

Martinif:
This adds additional notes if the meal is vegetarian or vegan.


klemens:
Thanks!

PR #108, opened 2019-11-08
niederbayern_oberpfalz: improve invalid newline detection

ghost:
Single invalid newlines weren't detected correctly. Improves fix of #107

### DIFF
--- a/parsers/niederbayern_oberpfalz.py
+++ b/parsers/niederbayern_oberpfalz.py
@@ -2,7 +2,7 @@
 #
 #  niederbayern_oberpfalz.py
 #
-#  Copyright 2016 Alex Flierl <shad0w73@vmail.me>
+#  Copyright 2019 Alex Flierl <shad0w73@freenet.de>
 #
 #  This program is free software: you can redistribute it and/or modify
 #  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


parsers/niederbayern_oberpfalz.py
```diff
@@ -61,6 +31,26 @@
 from pyopenmensa.feed import LazyBuilder
 
 
+# Only allow a newline after 8 separators
+def replace_invalid_newlines(s):
```
klemens: This can be fixed much simpler with the following regex, which matches one or more newlines, if they are not followed by the date pattern and semicolon (negative lookahead): `\n+(?!\d\d\.\d\d\.\d\d\d\d;)`

For reference, the problematic part looks like:
```
12.11.2019;Di;Suppe;Grießnockerlsuppe (3,A,AA,C,I);V;0,70 / 0,90 / 1,40;0,70;0,90;1,40
12.11.2019;Di;HG1;// Selbstbedienung //

Schinkennudeln (2,3,8,A,AA);S;2,00 / 2,80 / 3,60;2,00;2,80;3,60
12.11.2019;Di;HG2;Vegane Kartoffelpizza (3,F);VG;2,80 / 3,40 / 4,40;2,80;3,40;4,40
```

klemens (changes requested):
Have you tried reporting the problem upstream, i.e. at the Studentenwerk? Because this "csv file" is *really* broken.

PR #110, opened 2020-08-13
Add Mensa Mainz

kreativmonkey:
Adding the parser for the mensas of studierendenwerk-mainz.

### DIFF
--- a/parsers/mainz.py
+++ b/parsers/mainz.py
@@ -1,0 +1,201 @@
+from urllib.request import urlopen
+from bs4 import BeautifulSoup as parse
+from bs4.element import Tag
+import re
+import datetime
+import os.path as path
+
+from utils import Parser
+from pyopenmensa.feed import OpenMensaCanteen
+
+
+day_regex = re.compile('(?P<date>\d{4}-\d{2}-\d{2})')
+price_regex = re.compile('(?P<price>\d+[,.]\d{2}) ?€')
+notes_regex = re.compile('\[(?:(([A-Za-z0-9]+),?)+)\]$')
+extract_legend = re.compile('\((\w+,?)+\)')
+extract_legend_notes = re.compile('(?:([A-Za-z0-9]+))')
+extract_notes_regex = re.compile('(?:([A-Za-z0-9]+)[,|\)])')
+
+canteenLegend = {
+  # API Extraction: https://github.com/kreativmonkey/jgu-mainz-openmensa/issues/1
+  '0' : 'all',
+  '1' : 'Zentralmensa',
+  '2' : 'Mensa-Georg-Foster',
+  '3' : 'Cafe-Rewi',
+  '4' : 'Mensa-Bingen',
+  '5' : 'Mensa-K3',
+  '6' : 'Mensa-Holzstraße',
+  '7' : 'Mensarium',
+  '8' : 'Cafe-Bingen-Rochusberg',
+  '9' : 'Mensablitz'
+}
+
+display = {
+  '2' : 'Aktuelle Woche',
+  '3' : 'Nächste Woche'
+}
+
+roles = ('student', 'other', 'employee')
+
+extraLegend = {
+    # Source: https://www.studierendenwerk-mainz.de/essentrinken/speiseplan/
+    '1': 'mit Farbstoff',
+    '2': 'mit Konservierungsstoff',
+    '3': 'mit Antioxidationsmittel',
+    '4': 'mit Geschmacksverstärker',
+    '5': 'geschwefelt',
+    '6': 'geschwärzt',
+    '7': 'gewachst',
+    '8': 'Phosphat',
+    '9': 'mit Süßungsmitteln',
+    '10': 'enthält eine Phenylalaninquelle',
+    'S' : 'Schweinefleisch',
+    'G' : 'Geflügelfleisch',
+    'R' : 'Rindfleisch',
+    'Gl' : 'Gluten',
+    'We' : 'Weizen (inkl. Dinkel)',
+    'Ro' : 'Roggen',
+    'Ge' : 'Gerste',
+    'Haf': 'Hafer',
+    'Kr' : 'Krebstiere und Krebstiererzeugnisse',
+    'Ei' : 'Eier und Eiererzeugnisse',
+    'Fi' : 'Fisch und Fischerzeugnisse',
+    'En' : 'Erdnüsse und Erdnusserzeugnisse',
+    'So' : 'Soja und Sojaerzeugnisse',
+    'La' : 'Milch und Milcherzeugnisse',
+    'Sl' : 'Sellerie und Sellerieerzeugnisse',
+    'Sf' : 'Senf und Senferzeugnisse',
+    'Se' : 'Sesamsamen und Sesamsamenerzeugnisse',
+    'Sw' : 'Schwefeldioxid und Sulfite > 10mg/kg',
+    'Lu' : 'Lupine und Lupinerzeugnisse',
+    'Wt' : 'Weichtiere und Weichtiererzeugnisse',
+    'Nu' : 'Schalenfrüchte',
+    'Man': 'Mandel',
+    'Has': 'Haselnüsse',
+    'Wa' : 'Walnüsse',
+    'Ka' : 'Kaschunüsse',
+    'Pe' : 'Pecanüsse',
+    'Pa' : 'Paranüsse',
+    'Pi' : 'Pistatien',
+    'Mac': 'Macadamianüsse',
+    'icon:S.png' : 'Scheinefleisch',
+    'icon:R.png' : 'Rindfleisch',
+    'icon:Fi.png' : 'Fisch',
+    'icon:Gl.png' : 'Glutenfrei',
+    'icon:La.png' : 'Lactosefrei',
+    'icon:Vegan.png' : 'Vegan',
+    'icon:Veggi.png' : 'Vegetarisch'
+    
+}
+
+def build_meal_name(meal):
+	# Name des Gerichts
+	name = str(meal).strip()
+	# Remove the notes from Mealname and delete unnecessary spaces
+	name = ' '.join(re.sub(r'\((\w+,?)+\)', '', name).split())
+	if len(name) > 250:
+			name = name[:245] + '...' 
+	
+	return name
+	
+def extract_meal_notes(meal):
+	# extracting the legend
+	legpart = extract_legend.findall(str(meal).strip())
+	legend = []
+	for l in legpart:
+		legend.extend(extract_legend_notes.findall(l))
+	
+	notes = set()
+	for l in legend:
+		if extraLegend[l]:
+			notes.add(extraLegend[l])
+	
+	return notes
+	
+def build_meal_notes(meal):
+	notes = set()
+			
+	for icon in meal.find_all('img'):
+		#<img src="/fileadmin/templates/images/speiseplan/Fi.png"/>
+		#<img src="/fileadmin/templates/images/speiseplan/La.png"/>
+		if "icon:"+path.basename(icon['src']) in extraLegend:
+			notes.add(extraLegend["icon:"+path.basename(icon['src'])])
+			
+	# extracting the legend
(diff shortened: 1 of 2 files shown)


kreativmonkey:
There are a build issue but not from my changes, what should i do?

Y0hy0h:
You can do nothing. ;)

I haven't contributed here for a long time and don't have time to review right now. You'll have to wait until someone finds the time to review and merge, which usually is @klemens.

kreativmonkey:
No problem, I would like to fix the build problem, but at the moment I don't know enough about the project.

Y0hy0h:
As far as I can tell the problem is a 404 in the Darmstadt parser. (I'm not 100% certain, though.)

That means that it is unrelated to this PR. I think it would be best if it was fixed in a separate PR anyway. ;)

klemens:
Thank you for implementing this new parser!

I fixed darmstadt, but the review will have to wait until next week unfortunately.

kreativmonkey:
Thank you very much for the project, so I don't have to reinvent the wheel.

parsers/mainz.py
```diff
@@ -0,0 +1,177 @@
...
+  # The pricing for employee and others are the same!
+	meal_prices["student"] = prices[0].replace(',', '.')
+	meal_prices["employee"] = prices[1].replace(',', '.')
+	meal_prices["other"] = prices[1].replace(',', '.')
+	
+	return meal_prices
+	
+def parse_data(canteen, data):	
+	for v in data.find_all('div'):
+		if not v.has_attr('class'):
```
Y0hy0h, line 134: This algorithm implicitely assumes that the `div`s appear in a certain order. In case this order is violated in the future, I think it might be nice to have a comment here that reminds of that. Something like
```
# We assume that the `div`s appear in a certain order and will associate each meal to the previously encountered date and category.
```

parsers/mainz.py
```diff
@@ -0,0 +1,177 @@
...
+	return name
+	
+def build_meal_notes(meal):
+  meal_name = str(meal.find('div', class_="speiseplanname").string).strip()
+  images = meal.find_all('img')
+
+  # Use a set for easy elimination of duplicates
+  notes = set()
+      
+  # extracting the icons with spezial informations about the meal
```
Y0hy0h, line 109: Nit: Typo. Correction: `# Extracting the icons with special informations about the meal`

parsers/mainz.py
```diff
@@ -0,0 +1,177 @@
...
+    'Nu' : 'Schalenfrüchte',
+    'Man': 'Mandel',
+    'Has': 'Haselnüsse',
+    'Wa' : 'Walnüsse',
+    'Ka' : 'Kaschunüsse',
+    'Pe' : 'Pecanüsse',
+    'Pa' : 'Paranüsse',
+    'Pi' : 'Pistatien',
+    'Mac': 'Macadamianüsse',
+    'icon:S.png' : 'Scheinefleisch',
```
Y0hy0h, line 79: Consider splitting these off into their own dictionary, e.g. `iconLegend`, to avoid the `icon:` prefix. See line 112 for how this can simplify code.

parsers/mainz.py
```diff
@@ -0,0 +1,177 @@
...
+  meal_name = str(meal.find('div', class_="speiseplanname").string).strip()
+  images = meal.find_all('img')
+
+  # Use a set for easy elimination of duplicates
+  notes = set()
+      
+  # extracting the icons with spezial informations about the meal
+  # Example: <img src="/fileadmin/templates/images/speiseplan/Veggi.png"/>
+  for icon in images:
+    if "icon:"+path.basename(icon['src']) in extraLegend:
```
Y0hy0h, line 112: If you choose to split the icons into an extra dict, this could be simplified to
```
iconName = path.basename(icon['src'])
if iconName in iconLegend:
    notes.add(iconsLegend[iconName])
```

parsers/mainz.py
```diff
@@ -0,0 +1,177 @@
...
+      notes.add(extraLegend["icon:"+path.basename(icon['src'])])
+
+  for l in extract_legend_notes.findall(meal_name):
```
Y0hy0h, line 115: Nit: I prefer writing out variable names to make them easy to read. I suggest `extra` instead of `l`.

parsers/mainz.py
kreativmonkey: Thats a create idea. I've didn't saw another way to get the meals from the page.

parsers/mainz.py
```diff
@@ -0,0 +1,177 @@
...
+		  continue
+
+		if v['class'][0] == 'speiseplan_date':
+		  date = day_regex.findall(str(v.string).strip())[0]
+				  
+		if v['class'][0] == 'speiseplan_bldngall_name':
+		  canteen_name = str(v.string).strip()
+		  
+		if v['class'][0] == 'speiseplancounter':
+		  counter_name = str(v.string).strip()
```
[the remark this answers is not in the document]
kreativmonkey: I have renamed it as suggested and added a comment for better understanding.

parsers/mainz.py
```diff
@@ -36,68 +36,70 @@
...
+  'Wa' : 'Walnüsse',
+  'Ka' : 'Kaschunüsse',
+  'Pe' : 'Pecanüsse',
+  'Pa' : 'Paranüsse',
+  'Pi' : 'Pistatien',
+  'Mac': 'Macadamianüsse',
+}
+
+iconLegend = {
+  'icon:S.png' : 'Scheinefleisch',
```
Y0hy0h: I think these now need to lose the `icon:` prefix, since in line 114 you look up just the `basename`.

Y0hy0h: Oh, that was quick. 😂

parsers/mainz.py
```diff
@@ -0,0 +1,182 @@
...
+import os.path as path
+
+from utils import Parser
+from pyopenmensa.feed import OpenMensaCanteen
+
+
+day_regex = re.compile(r'(\d{2}.\s\w+\s\d{4})')
+price_regex = re.compile(r'(?P<price>\d+[,.]\d{2}) ?€')
+notes_regex = re.compile(r'\[(?:(([A-Za-z0-9]+),?)+)\]$')
+extract_legend = re.compile(r'\((\w+,?)+\)')
```
klemens, line 14: The above two are not used?

klemens (changes requested):
Thanks again for writing this parser! This looks quite good already. But you could use more of the build-in functionality of the LazyBuilder/OpenMensaCanteen to avoid re-implementing the same functionality. :wink:

You can add your parser to `build_scripts/test-urls.txt` so it gets tested in CI.

Also thanks @Y0hy0h for your review!

parsers/mainz.py
```diff
@@ -0,0 +1,182 @@
...
+extract_legend_notes = re.compile(r'(?<=[\(,])(\w{1,2})')
+
+canteens = {
+  # API Extraction: https://github.com/kreativmonkey/jgu-mainz-openmensa/issues/1
+  '0' : 'all',
```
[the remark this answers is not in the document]
kreativmonkey: You're right, i remove "all"

parsers/mainz.py
```diff
@@ -0,0 +1,182 @@
...
+      notes.add(iconLegend[icon_name])
+
+  for extra in extract_legend_notes.findall(meal_name):
+    if extra in extraLegend:
+      notes.add(extraLegend[extra])
+
+  return list(notes)
+	
+def build_meal_price(meal):
+	meal_prices = {}
```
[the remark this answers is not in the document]
kreativmonkey: normaly there should be two spaces, i don't know whats going wrong but i've fixed it.

parsers/mainz.py
```diff
@@ -0,0 +1,182 @@
...
+  name = ' '.join(re.sub(r'\((\w+,?)+\)', '', str(meal)).split())
+
+  # Shorten the meal name to 250 characters like the api specification: https://doc.openmensa.org/feed/v2/#name
+  if len(name) > 250:
+    name = name[:245] + '...' 
+
+  return name
+	
+def build_meal_notes(meal):
+  meal_name = str(meal.find('div', class_="speiseplanname").string).strip()
```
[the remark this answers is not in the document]
kreativmonkey: The notes are in the meal name and that is why i extract the name twice. The first one is to get a beautiful meal name without extras and the second is to get only the extras. I'm open for better ideas.

parsers/mainz.py
```diff
@@ -0,0 +1,182 @@
...
+    if v['class'][0] == 'speiseplancounter':
+      # Save the countername as category to list meals by counter
+      category = str(v.string).strip()
+      
+    if v['class'][0] == 'menuspeise':
+      meal_name = build_meal_name(v.find('div', class_="speiseplanname").string)
+      meal_notes = build_meal_notes(v)
+      meal_prices = build_meal_price(v)
+
+      canteen.addMeal(date, category,
```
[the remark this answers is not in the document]
kreativmonkey: I can add it but i don't think that the date can be None if there is a menuspeise class on the page.

parsers/mainz.py
```diff
@@ -0,0 +1,182 @@
...
+			speiseplan = resp.find('div', class_='speiseplan')
+		
+		canteen = parse_data(canteen, speiseplan)
+
+
+	return canteen.toXMLFeed()
+
+parser = Parser('mainz',
+				handler=parse_url,
+				shared_prefix='https://www.studierendenwerk-mainz.de/speiseplan/frontend/index.php')
```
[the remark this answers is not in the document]
kreativmonkey: I use the URL from which the website itself gets its data (extracted with the dev-tools). This page only contains the necessary data. If there should be a change, the structure of the page will surely change, so I think this way is the best way.

parsers/mainz.py
```diff
@@ -0,0 +1,182 @@
...
+def parse_url(url, today=False):
+	canteen = OpenMensaCanteen()
+  
+  # There are two displays one for the current and one for the next week
+	for d in display:
+		with urlopen(url + '&display_type=' + d) as resp:
+			resp = parse(resp.read().decode('utf-8', errors='ignore'), features='lxml')
+			speiseplan = resp.find('div', class_='speiseplan')
+		
+		canteen = parse_data(canteen, speiseplan)
```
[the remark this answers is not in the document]
kreativmonkey: thats to extend the canteen with the data from the second week. it's like 

```
i = 0
i = i + 1
```
I can't think of a better solution right now.

parsers/mainz.py
```diff
@@ -0,0 +1,182 @@
...
+      meal_notes = build_meal_notes(v)
+      meal_prices = build_meal_price(v)
+
+      canteen.addMeal(date, category,
+              meal_name, meal_notes, meal_prices)
+	
+  return canteen
+
+
+def parse_url(url, today=False):
```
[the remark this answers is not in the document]
kreativmonkey: I implemented the today-feed.

kreativmonkey:
Sorry for the delay, I had a lot to do in the last weeks. Now I have adapted everything according to Kalmens wishes. I can understand that you want to avoid superfluous code, I just wanted to simplify the work in the future in case you decide to add the "all" mensas option.

parsers/mainz.py
```diff
@@ -124,13 +124,10 @@ def parse_data(canteen, data):
     if 'speiseplan_bldngall_name' in v['class']:
       canteen_name = str(v.string).strip()
       
-    if 'speiseplancounter' in v['class']:
-      # Save the countername as category to list meals by counter
-      category = str(v.string).strip()
```
klemens: I think you meant to remove the if above that one? 😅 Otherwise category will always be None.

klemens (changes requested):
No problem, but I think you made a slip?

parsers/mainz.py
kreativmonkey: Oh no, F****! I'm so sorry, correction is on the way.

kreativmonkey:
I think i can't do anything for the failed travis job. There was a problem while downloading the build_scripts/test_urls.txt but they are still there!

Y0hy0h (approved):
I glanced over the code and it looks good to me. Only a small suggestion. :)

Issue #113, opened 2021-04-15
Thuringia updated


k-0:
The site from canteens in thuringia was updated. Openmensa doesn't show meals anymore.

PR #118, opened 2021-08-02
Thueringen redesign

hermlon:
I rewrote the parser for thueringen in response to a complete redesign of their website (mentioned in  #113 ). This is my first time contributing to openmensa, so I'm not sure whether I got everything correct.

I left out the canteen metadata parsing, because the canteen information page on, for example `https://www.stw-thueringen.de/mensen/erfurt/mensa-nordh%C3%A4user-stra%C3%9Fe.html` isn't easily derived from the corresponding `canteen_id` `44`. Also the only information which is likely to change, the opening times, aren't provided in an easily machine readable format.

Also `we-horn` doesn't seem to exist anymore, so I removed it.

Since meal categories don't really exist, I chose "Vegan", "Vegetarisch", "Fisch" and everything else will be listed under "Fleisch".

I'd be glad for any feedback and look forward to having the meals available in the openmensa app again.

### DIFF
--- a/parsers/thueringen.py
+++ b/parsers/thueringen.py
@@ -6,106 +6,13 @@ from datetime import datetime, timedelta, date
 from pyopenmensa.feed import LazyBuilder
 import ssl
 
-#
-#
-#def parse_fees(document):
-#	fees_regex = re.compile('Bedienstete.*')
-#
-#	fees = document.find_all('p', class_='MsoNormal', string=fees_regex)
-#
-#	for fee in fees:
-#		fee_strings = fees_regex.findall(fee.text)
-#		for amount_candidate in fee_strings:
-#			amount_strings = amount_regex.findall(amount_candidate)
-#			if len(amount_strings) != 2:
-#				continue
-#
-#			employees_fee = float(amount_strings[0].replace(',', '.'))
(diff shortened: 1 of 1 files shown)


mswart:
Thx for the change. I don't have the time/capacity to actually maintain these anymore. As it is currently broken, this is clearly an improvement.
With the merge the new version should get online tomorrow or so. If this is not the case, ask again. If you want direct feedback from me, it is usually best to ping me (via email or in the #openmensa irc channel on freenode).

hermlon:
Hi, to me it still looks like the update wasn't applied yet. At least there is no information on the openmensa.org page. Could you maybe have a look at it again?

klemens:
@hermlon Sorry for that, I changed the url of the Thueringen parser to my server previously, back when @mswart's server wasn#t working correctly. I have now also updated my server and refreshed the canteens, seems to work fine now! 🙂

However, it seems je-eah and sk-mensa are not working correctly.

hermlon:
@klemens thanks a lot, I added checks for the errors which occurred to you and created another pull request, I wasn't careful enough and assumed elements to always exist (apparently there are even meals without names - but *with* allergens and additives).

PR #119, opened 2021-09-27
Thueringen empty elements fix

hermlon:
This fixes the errors mentioned here https://github.com/mswart/openmensa-parsers/pull/118#issuecomment-927316810 by checking for the existence of meal names, additives, allergens and prices. I also removed an apparently now unnecessary ssl workaround.


klemens:
Thanks, merged and pushed to my server!

PR #134, opened 2022-04-19
update bs4-path to legend & additions for muenchen-parser

andiarbeit12:


moritzsternemann:
Thanks for the fixes! Is there anything I can do to help get this merged @mswart?

There is also #130 which seems be achieving the same goal as this PR...

PR #136, opened 2022-06-27
Updated parser to work with new website

henninglulei:


blitzdose:
Hi, please merge this PR as soon as possible. We developed an App for our university and are really depending on the openmensa-data. But with this parser not working at the moment for Karlsruhe, the data don't get updated.

